PR #10, opened 2022-07-19
Add tests for updating token price

AStox:


src/Connector.sol
```diff
@@ -105,7 +105,12 @@ contract CentrifugeConnector is Test {
         bytes16 trancheId,
         uint256 price
     ) public onlyRouter {
+        Pool storage pool = pools[poolId];
+        require(pool.createdAt > 0, "CentrifugeConnector/invalid-pool");
+
         Tranche storage tranche = tranches[poolId][trancheId];
+        require(tranche.latestPrice > 0, "CentrifugeConnector/invalid-tranche");
+
```
offerijns, line 113: The tranche can only exist if the pool exists, so to prevent 1 storage read, I think we can do:

```suggestion
        Tranche storage tranche = tranches[poolId][trancheId];
        require(tranche.latestPrice > 0, "CentrifugeConnector/invalid-pool-or-tranche");

```

offerijns (approved):

PR #13, opened 2022-07-20
Add fail-case tests for updateMember and updateTokenPrice

AStox:


src/Connector.sol
```diff
@@ -116,6 +117,8 @@ contract CentrifugeConnector is Test {
         address user,
         uint256 validUntil
     ) public onlyRouter {
+        Tranche storage tranche = tranches[poolId][trancheId];
+        require(tranche.latestPrice > 0, "CentrifugeConnector/invalid-pool-or-tranche");
         RestrictedTokenLike token = RestrictedTokenLike(tranches[poolId][trancheId].token);
```
offerijns, line 122: There's a duplicate SLOAD here:
```suggestion
        RestrictedTokenLike token = RestrictedTokenLike(tranche.token);
```

offerijns (approved):

PR #14, opened 2022-07-22
add depost & withdraw functions

ilinzweilin:


src/Messages.sol
```diff
@@ -13,7 +13,10 @@ library ConnectorMessages {
         AddTranche,
         UpdateTokenPrice,
         UpdateMember,
-        TransferTo
+        TransferTo, 
+        Deposit,
```
offerijns, line 17: Maybe I'm thinking about this wrong, but I don't think we need separate messages. I think we just need a single `Transfer` message, with arguments `pool_id`, `tranche_id`, `destination_domain`, and `destination_address`.

E.g, a user call `deposit` on domain A, to initiate a transfer. This burns the tokens and sends a `Transfer` message to the `Centrifuge` domain (since all transfers go through centrifuge chain). If this goes to domain B, then on domain B the `Transfer` message is received and this mints the tokens here. So the `deposit` method would be public and calls the router to send the message, the `withdraw` method is not public and can only be called by the router to mint the tokens on the destination chain.

PR #16, opened 2022-07-22
Connectors v1

ilinzweilin:
### To Do

- [x] Use cent-chain generated messages for all encoding/decoding tests in `Messages.t.sol` @NunoAlexandre 
- [x] only allow domainId == Cent Chain - https://github.com/centrifuge/connectors/pull/16#discussion_r1087620699 @AStox 
- [x] Add `domain` to the `Transfer` message - see https://github.com/centrifuge/connectors/pull/16#discussion_r950462971 @AStox 
- [x] Remove domainLookup - https://github.com/centrifuge/connectors/pull/16#discussion_r1084196068 @AStox 
- [x] Testing that a transfer to a non-whitelisted address fails - https://github.com/centrifuge/connectors/pull/16#discussion_r1084204578 @AStox 
- [x] Check on line 116-11 - https://github.com/centrifuge/connectors/pull/16#discussion_r1084202876 @AStox 
- [x] Use cent-chain generated transfer message in encoding/decoding tests - @NunoAlexandre


src/Connector.sol
```diff
@@ -135,18 +135,19 @@ contract CentrifugeConnector {
         memberlist.updateMember(user, validUntil);
     }
 
-    function deposit(
```
AStox: Okay I know I already asked for these to be renamed once, but I wonder if it wouldn't be clearer to name them `transferIn`, and `transferOut`. I worry `transfer` and `transferTo` will be confusing. What do you think?

test/Connector.t.sol
```diff
@@ -180,71 +182,290 @@ contract ConnectorTest is Test {
...
-    // function testDepositFromOtherOriginFails(uint64 poolId) public { }
+    function testTransferForNonExistentPoolFails(uint64 poolId, bytes16 trancheId, address user, uint256 amount, uint256 validUntil) public { 
+        vm.assume(validUntil > safeAdd(block.timestamp, minimumDelay));
+
+        // do not add pool
+        vm.expectRevert(bytes("CentrifugeConnector/invalid-pool"));
+        homeConnector.addTranche(poolId, trancheId, "Some Name", "SYMBOL");
+        
+        vm.expectRevert(bytes("CentrifugeConnector/invalid-pool-or-tranche"));
+        homeConnector.updateMember(poolId, trancheId, user, validUntil);
```
AStox, line 203: Is this run after the first revert in this test? I assumed the test ends after the first revert.

test/Connector.t.sol
```diff
@@ -180,71 +182,290 @@ contract ConnectorTest is Test {
...
+        homeConnector.updateMember(poolId, trancheId, user, uint(-1));
+        homeConnector.transfer(poolId, trancheId, user, amount);
+        
+        // approve token
+        (address token_,,) = bridgedConnector.tranches(poolId, trancheId);
+        RestrictedTokenLike token = RestrictedTokenLike(token_);
+        token.approve(address(bridgedConnector), uint(-1)); // approve connector to take token
+
+        // transferTo
+        vm.expectRevert(bytes("!remote"));
```
AStox, line 338: This is an odd revert message. Is there no way to make this more verbose?

test/Connector.t.sol
```diff
@@ -180,71 +182,290 @@ contract ConnectorTest is Test {
...
+        uint32 domainId = 3000;
+        bytes32 recipient = stringToBytes32("0xefc56627233b02ea95bae7e19f648d7dcd5bb132");
+        user = address(this); // set deployer as user to approve the cnnector to transfer funds
+
+        // add Centrifuge domain to router                  
+        bridgedRouter.enrollRemoteRouter(domainId, recipient);
+        
+        // add Centrifuge domain to connector
+        assertEq(bridgedConnector.wards(address(this)), 1);
+        bridgedConnector.file("domain", domainName, domainId); // use wrong domainID
```
AStox, line 366: Can remove the `// use wrong domainID` comment

test/Connector.t.sol
```diff
@@ -180,71 +182,290 @@ contract ConnectorTest is Test {
...
```
AStox, line 396: Can remove the `// use wrong domainID` comment

AStox (commented):
Looks good, but some of the comments in the tests are misleading or inaccurate.

src/Connector.sol
```diff
@@ -152,7 +152,7 @@ contract CentrifugeConnector {
         address user,
         uint256 amount,
         string memory domainName
-    ) public auth {
```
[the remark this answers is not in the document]
ilinzweilin: we decided to go for public :)

src/Connector.sol
ilinzweilin: sure, open to rename

test/Connector.t.sol
ilinzweilin: the test continues and passes in case all the reverts were expected

test/Connector.t.sol
ilinzweilin: the Nomad Router contract that we are inheriting from is throwing that one

src/Connector.sol
```diff
@@ -124,15 +135,33 @@ contract CentrifugeConnector is Test {
         memberlist.updateMember(user, validUntil);
     }
 
-    function transferTo(
+    function transfer(
```
AStox, line 138: Rename to `transferIn`

src/Connector.sol
```diff
@@ -124,15 +135,33 @@ contract CentrifugeConnector is Test {
...
         uint256 amount
     ) public onlyRouter {
         RestrictedTokenLike token = RestrictedTokenLike(tranches[poolId][trancheId].token);
+        require(address(token) != address(0), "CentrifugeConnector/unknown-token");
         require(token.hasMember(user), "CentrifugeConnector/not-a-member");
         token.mint(user, amount);
     }
-    
+
+    function transferTo(
```
AStox, line 150: rename to `transferOut`

src/Connector.sol
offerijns: I would strongly suggest going with `transfer` for the public method, the internal method can be anything (something like `handleTransfer` / `onTransfer` / `executeTransfer`?). Reason being: the internal method won't be called by anyone externally, it's just called by the router. So only the name of the external public method really matters, and in this case the intuitive name imo is just `transfer`.

src/Connector.sol
```diff
@@ -96,7 +107,7 @@ contract CentrifugeConnector is Test {
 
         address memberlist = memberlistFactory.newMemberlist();
         RestrictedTokenLike(token).depend("memberlist", memberlist);
-
+        MemberlistLike(memberlist).updateMember(address(this), uint(-1)); // required to be able to receive tokens in case of withdrawals
```
offerijns, line 110: 💯

src/Messages.sol
```diff
@@ -135,4 +135,29 @@ library ConnectorMessages {
...
+     */
+    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount) internal pure returns (bytes memory) {
+        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, amount);
+    }
+
+    function isTransfer(bytes29 _msg) internal pure returns (bool) {
+        return messageType(_msg) == Call.Transfer;
+    }
+
+    function parseTransfer(bytes29 _msg) internal pure returns (uint64 poolId, bytes16 trancheId, address user, uint256 amount) {
```
offerijns, line 156: I think the transfer message should probably contain the domain. For now, transfers will only go from a connector to cent chain and vice verse. But in the future, transfers from one domain to another (non cent chain one) will be possible, but they will always still go through cent chain. So cent chain will need to know the destination domain in order to forward it correctly.

src/routers/xcm/Router.sol
```diff
@@ -48,13 +59,38 @@ contract ConnectorXCMRouter is Router, Test {
...
+            (uint64 poolId, bytes16 trancheId, address user, uint256 amount) = ConnectorMessages.parseTransfer(_msg);
+            connector.transfer(poolId, trancheId, user, amount);
         } else {
             require(false, "invalid-message");
         }
     }
+
+    function sendMessage(uint32 destinationDomain, uint64 poolId, bytes16 trancheId, uint256 amount, address user) external onlyConnector {
+        bytes32 remoteAddress = _mustHaveRemote(destinationDomain);
+        Home(xAppConnectionManager.home()).dispatch(
```
offerijns, line 77: This is the Nomad logic that's now added to the XCM router. For XCM, we don't know exactly yet how `sendMessage` will work. Most likely there will be some precompiled contract that we'll need to call.

offerijns (commented):
Really thorough testing, nice :)

src/Connector.sol
AStox: Good point! Agreed 👍

src/Connector.sol
```diff
@@ -78,6 +81,14 @@ contract CentrifugeConnector is Test {
         emit File(what, data);
     }
 
+    function file(bytes32 name, string memory domainName, uint32 domainId) public auth  {
```
NunoAlexandre, line 84: It seems uncommon to document Solidity code? For someone new to a codebase like this, I can't tell why we would need this function and where it would be used. Can we document that or explore that here in the comments?

src/Connector.sol
```diff
@@ -111,7 +122,7 @@ contract CentrifugeConnector is Test {
 
         address memberlist = memberlistFactory.newMemberlist();
         RestrictedTokenLike(token).depend("memberlist", memberlist);
-
+        MemberlistLike(memberlist).updateMember(address(this), uint(-1)); // required to be able to receive tokens in case of withdrawals
```
NunoAlexandre, line 125: I am a bit confused about this call; the function definition I find is this below so how are we passing the address and a negative `validUntil`? Also, does `-1` `validUntil` means we set it as valid indefinitely? 

```solidity
uint64 poolId,
bytes16 trancheId,
address user,
uint256 validUntil
```

src/Connector.sol
```diff
@@ -139,15 +150,33 @@ contract CentrifugeConnector is Test {
...
 
-    function transferTo(
+    function handleTransfer(
         uint64 poolId,
         bytes16 trancheId,
         address user,
         uint256 amount
     ) public onlyRouter {
         RestrictedTokenLike token = RestrictedTokenLike(tranches[poolId][trancheId].token);
+        require(address(token) != address(0), "CentrifugeConnector/unknown-token");
```
NunoAlexandre, line 160: Why `require(address(token) != address(0)` ?

src/Connector.sol
```diff
@@ -139,15 +150,33 @@ contract CentrifugeConnector is Test {
...
-    
+
+    function transfer(
+        uint64 poolId,
+        bytes16 trancheId,
+        address user,
+        uint256 amount,
+        string memory domainName
+    ) public {
+        uint32 domainId = domainLookup[keccak256(bytes(domainName))];
```
NunoAlexandre, line 172: Is it an idea that we create a function to lookup a domain given a domain name? To avoid duplicating this lookup impl logic every time we look it up

src/Messages.sol
NunoAlexandre: It seems that this is still missing the `domain` here as @offerijns mentioned

src/routers/xcm/Router.sol
```diff
@@ -43,13 +50,34 @@ contract ConnectorXCMRouter {
...
             connector.addTranche(poolId, trancheId, tokenName, tokenSymbol);
         } else if (ConnectorMessages.isUpdateMember(_msg) == true) {
-            (uint64 poolId, bytes16 trancheId, address user, uint256 amount) = ConnectorMessages.parseUpdateMember(_msg);
-            connector.updateMember(poolId, trancheId, user, amount);
+            (uint64 poolId, bytes16 trancheId, address user, uint256 validUntil) = ConnectorMessages.parseUpdateMember(_msg);
+            connector.updateMember(poolId, trancheId, user, validUntil);
         } else if (ConnectorMessages.isUpdateTokenPrice(_msg) == true) {
             (uint64 poolId, bytes16 trancheId, uint256 price) = ConnectorMessages.parseUpdateTokenPrice(_msg);
             connector.updateTokenPrice(poolId, trancheId, price);
+        } else if (ConnectorMessages.isTransfer(_msg) == true) {
```
NunoAlexandre, line 58: Is this a solidity thing, the need to compare a boolean value explicitly against a literal boolean? When I was a student we would get punished for doing this 😆 

```suggestion
        } else if (ConnectorMessages.isTransfer(_msg)) {
```

src/routers/xcm/Router.sol
```diff
@@ -43,13 +50,34 @@ contract ConnectorXCMRouter {
...
         } else {
             require(false, "invalid-message");
         }
     }
+
+    function sendMessage(uint32 destinationDomain, uint64 poolId, bytes16 trancheId, uint256 amount, address user) external onlyConnector {
+        // TODO: implement
+    }
+
+    function bytes32ToString(bytes32 _bytes32) internal returns (string memory) {
```
NunoAlexandre, line 70: We should dry this but for now let's leave a `TODO` comment to do so later

src/routers/xcm/Router.sol
```diff
@@ -43,13 +50,34 @@ contract ConnectorXCMRouter {
...
+        } else if (ConnectorMessages.isTransfer(_msg) == true) {
+            (uint64 poolId, bytes16 trancheId, address user, uint256 amount) = ConnectorMessages.parseTransfer(_msg);
+            connector.handleTransfer(poolId, trancheId, user, amount);
         } else {
             require(false, "invalid-message");
         }
     }
+
+    function sendMessage(uint32 destinationDomain, uint64 poolId, bytes16 trancheId, uint256 amount, address user) external onlyConnector {
+        // TODO: implement
```
NunoAlexandre, line 67: I am curious as to how this function will look like in the future. Will it require a specific impl per domain?

NunoAlexandre (commented):
Got some questions 🤓

src/Connector.sol
offerijns: `file` is just a Maker standard name for an "updating parameters" method.

src/Connector.sol
offerijns: `-1` is basically a max integer, so whitelisting for an infinite period. @AStox you can also do `type(uint256).max` which is more readable.

I'm not sure where you found the function definition. If you look at the `MemberlistLike` interface, the interface is
```
function updateMember(address usr, uint validUntil) external
```

The pool id and tranche id aren't needed because a memberlist contract is deployed for each specific token, so it's already implicit.

src/Connector.sol
offerijns: You can't transfer tokens if the token contract not exist yet on this domain. In solidity, there's no options/nullable variables, rather everything is defaulted to it's zero-like value. So address(0) is what a map to an address will be set to by default, indicating "empty".

src/Connector.sol
```diff
@@ -78,6 +81,14 @@ contract CentrifugeConnector is Test {
...
+        if(name == "domain") {
+           domainLookup[keccak256(bytes(domainName))] = domainId;
+           emit File(name, domainName);
+        } else { revert ("unknown name");}
+        
+    }
+
     // --- Internal ---
     function addPool(uint64 poolId) public onlyRouter {
         Pool storage pool = pools[poolId];
```
offerijns, line 94: I can't leave a comment there, but I just realised line 108, setting the token price to 1.0 is a rather bad design. There is no guarantee that a token is added to another domain at this price.

I think there are two alternatives:
1. We can't initialize to `0` as a token can actually get a price of 0, so we initialize it to `type(uint256).max)`. Downside is that this will need to be accounted for in external integrations, so they don't accidentally use this for calculations and quickly overflow.
2. We pass the current token price when calling `addTranche`. @NunoAlexandre would be good to get your input here, but this seems viable to me, and would cleanly initiate the tranche on the domain with the right values.

src/Connector.sol
```diff
@@ -111,7 +122,7 @@ contract CentrifugeConnector is Test {
```
offerijns, line 122: I wonder whether we actually need the check on line 116-117. It adds an `SLOAD` for the pool storage. But afaik through the check in `addTranche`, the value `tranches[poolId][trancheId]` cannot exist if the `poolId` does not already exist.

src/Connector.sol
```diff
@@ -139,15 +150,33 @@ contract CentrifugeConnector is Test {
...
-    function transferTo(
+    function handleTransfer(
         uint64 poolId,
         bytes16 trancheId,
         address user,
         uint256 amount
     ) public onlyRouter {
         RestrictedTokenLike token = RestrictedTokenLike(tranches[poolId][trancheId].token);
+        require(address(token) != address(0), "CentrifugeConnector/unknown-token");
         require(token.hasMember(user), "CentrifugeConnector/not-a-member");
```
offerijns, line 161: Do we have a test for this already, i.e. testing that a transfer to a non-whitelisted address fails? I think this check is actually redundant as the underlying erc20 token already checks that the destination address is a member.

src/Connector.sol
NunoAlexandre: > The pool id and tranche id aren't needed because a memberlist contract is deployed for each specific token, so it's already implicit.

I see. The definition I was referring to is defined in `src/Connector.sol` but I get it that these are at different levels of abstraction 👍

src/Connector.sol
NunoAlexandre: oh I see, it's like ensuring the address of the token `!= null`. Thanks!

src/routers/xcm/Router.sol
NunoAlexandre: That makes sense. I guess if we would integrate with another xcm-based domain we would need to handle that here.

src/Connector.sol
NunoAlexandre: > We pass the current token price when calling addTranche. @NunoAlexandre would be good to get your input here, but this seems viable to me, and would cleanly initiate the tranche on the domain with the right values.


Sounds good to me yeah 👍

src/Connector.sol
offerijns, line 172: One thing I just realised: we should only allow `domainId == Cent Chain` here. Transfers from one connector domain to another connector domain will get much more complex, and is just something we should figure out later.

src/Connector.sol
AStox: Not sure I follow... Unless it was changed, the check I believe you're talking about is this one:
```
Pool storage pool = pools[poolId];
require(pool.createdAt > 0, "CentrifugeConnector/invalid-pool");
```

Just checking that the pool exists, not the tranche.

src/Connector.sol
NunoAlexandre: Done ✅

src/Messages.sol
AStox: Then I think it should be `formatTransfer` that adds the domain right? Since it's on the receiving end, `parseTransfer` shouldn't care about the destination domain, since it's already arrived and been validated by cent-chain.

AStox (dismissed):

AStox (dismissed):

src/Connector.sol
```diff
@@ -5,12 +5,13 @@ pragma abicoder v2;
 import { RestrictedTokenFactoryLike, MemberlistFactoryLike } from "./token/factory.sol";
 import { RestrictedTokenLike } from "./token/restricted.sol";
 import { MemberlistLike } from "./token/memberlist.sol";
-import "forge-std/Test.sol";
+import { ConnectorMessages } from "src/Messages.sol";
```
offerijns, line 8: Can we add a TODO here or as an issue to change this again in V2?

src/Connector.sol
offerijns: @AStox In `deployTranche`, we are checking first if the pool exists, and then if the tranche exists. I think we don't need both checks, if the tranche exists then the pool also must exist.

src/Connector.sol
offerijns: Bumping this ^ @NunoAlexandre @AStox

src/Connector.sol
AStox: Actually it looks like we're only checking if the pool exists. That should probably be just checking if the tranche exists instead.

src/Connector.sol
AStox: Minting doesn't check the memberlist. This was the change I had made to `restricted.sol`, since it seems kind of odd that a permissioned token can mint to any address

offerijns: Gotcha. I would keep it like this for now, to not change the existing restricted token implementation. But let's discuss this a bit more and we can still change it in V2 optionally.

NunoAlexandre (approved):
It's happening 🌟 🚀

offerijns (approved):
LGTM! 💯

PR #19, opened 2022-07-22
Fix failing tests

### LINKED ISSUE #20: Add `testUpdatingMemberBeforeMinimumDelayFails`
We should add specific tests `testUpdatingMemberBeforeMinimumDelayFails` that check that updating a member to `< now + minimumDelay` reverts.

_Originally posted by @offerijns in https://github.com/centrifuge/connectors/pull/19#discussion_r928097596_

### LINKED ISSUE #21: Fix failing tests
Fix all failing tests

AStox:
Closes #20 
Closes #21

### DIFF
--- a/test/Connector.t.sol
+++ b/test/Connector.t.sol
@@ -97,6 +97,16 @@ contract ConnectorTest is Test {
         assertEq(memberlist.members(user), validUntil);
     }
 
+    function testUpdatingMemberBeforeMinimumDelayFails(uint64 poolId, bytes16 trancheId, address user, uint256 validUntil) public {
+        vm.assume(validUntil < safeAdd(block.timestamp, new Memberlist().minimumDelay()));
+        vm.assume(user != address(0));
+
+        homeConnector.addPool(poolId);
+        homeConnector.addTranche(poolId, trancheId, "Some Name", "SYMBOL");
+        vm.expectRevert("invalid-validUntil");
+        homeConnector.updateMember(poolId, trancheId, user, validUntil);
+    }
+
     function testUpdatingMemberAsNonRouterFails(uint64 poolId, bytes16 trancheId, address user, uint256 validUntil) public {
         vm.assume(validUntil > block.timestamp);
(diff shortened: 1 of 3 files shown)


test/Connector.t.sol
```diff
@@ -79,7 +79,7 @@ contract ConnectorTest is Test {
     }
 
     function testUpdatingMemberWorks(uint64 poolId, bytes16 trancheId, address user, uint256 validUntil) public {
-        vm.assume(validUntil > block.timestamp);
+        vm.assume(validUntil > safeAdd(block.timestamp, new Memberlist().minimumDelay()));
```
offerijns, line 82: We should add specific tests `testUpdatingMemberBeforeMinimumDelayFails` that check that updating a member to `< now + minimumDelay` reverts.

test/Connector.t.sol
```diff
@@ -50,8 +50,8 @@ contract ConnectorTest is Test {
         assertTrue(token_ != address(0));
 
         RestrictedTokenLike token = RestrictedTokenLike(token_);
-        assertEq(token.name(), tokenName);
-        assertEq(token.symbol(), tokenSymbol);
+        assertEq(token.name(), bytes32ToString(stringToBytes32(tokenName)));
```
offerijns, line 53: Can you add a comment explaining why this is necessary?

AStox: Sure! Actually I spent a lot of time thinking about this one... I ended up on the fact that probably the current behaviour is correct (where a string of "\0\0\0\0" gets treated as an empty string), so the problem then was that the test didn't reflect that. Do you agree?

I'll add a comment explaining whats going on here though

test/Connector.t.sol
AStox: Good idea

offerijns: https://github.com/centrifuge/connectors/issues/20

offerijns (approved):

offerijns (approved):

PR #27, opened 2022-07-31
update deploy script and add test

AStox:


AStox:
Successfully deploys:
1. RestrictedTokenFactory @ 0x0x936e7043f204cd36cd92009d44bf6b8129f6007b
2. MemberlistFactory @ 0x0x3d7c0f2542091881d6ca24b30f84e2ea9478ea0f
3. CentrifugeConnector @ 0x0xaecfa11fe9601c1b960661d7083a08a5df7c1947
4. ConnectorXCMRouter @ 0x0xe7986cc47fa7ef36763848317e6f92d1e8ac5fe7 

Still can't figure out how to correctly call the ConnectorXCMRouter's `handle()` function using `cast send`.

Also unable to verify any of these contracts on moonscan. Seems like foundry's `forge verify-contract` command is only for etherscan. Flattening the contracts and verifying via moonscan's verifcation UI not working either.

script/Connector-XCM.s.sol
```diff
@@ -11,14 +11,14 @@ contract ConnectorXCMScript is Script {
...
-        vm.broadcast();
+        vm.startBroadcast();
 
         address tokenFactory_ = address(new RestrictedTokenFactory());
         address memberlistFactory_ = address(new MemberlistFactory());
         CentrifugeConnector connector = new CentrifugeConnector(tokenFactory_, memberlistFactory_);
 
-        // TODO: add centrifugeChainOrigin_ arg
-        ConnectorXCMRouter router = new ConnectorXCMRouter(address(connector), address(0));
+        ConnectorXCMRouter router = new ConnectorXCMRouter(address(connector), address(0x7369626CEF070000000000000000000000000000));
```
offerijns, line 20:
```suggestion
        ConnectorXCMRouter router = new ConnectorXCMRouter(address(connector), address(vm.envAddress("CENTRIFUGE_CHAIN_ORIGIN")));
```

src/routers/xcm/Router.sol
```diff
@@ -34,12 +34,31 @@ contract ConnectorXCMRouter is Router, Test {
         _;
     }
 
+    // TODO: re-add onlyCentrifugeChainOrigin
```
offerijns, line 37: I think we found out the authentication was not the problem and we can remove this `handleUnauthenticated` method again?

src/routers/xcm/Router.sol
```diff
@@ -57,4 +76,14 @@ contract ConnectorXCMRouter is Router, Test {
             require(false, "invalid-message");
         }
     }
+
+    // TODO: For testing purposes. Remove once router integration is complete
```
offerijns, line 80: Likewise, I think these two methods can be removed again?

NunoAlexandre (approved):

PR #29, opened 2022-08-09
add Utils contract

AStox:


offerijns:
In my opinion:
- Whenever it doesn't add too much bloat, in non-test code the utils should be inlined rather than separated into a library. This leads to smart contracts that are entirely self-contained, without any imports, which eases auditing and reviewing.
   => `safeAdd` (and other math methods) in `memberlist.sol` should remain inlined.

- Test utils should be separated from non-test utils. We will likely add more test utils sooner rather than later, and combining these into a single library will lead to having more and more utils in the non-test code, which is also not ideal from an auditing perspective.
   => I would recommend creating two util files for now, `src/utils/MessageUtils.sol` with encoding/decoding utils for `Messages`, and `test/utils/TestUtils.sol` for the various test contracts.

AStox:
Thanks for the feedback @offerijns. Updated the utils contract to match your suggestions.

offerijns:
@AStox we should probably close this and start from scratch after merging https://github.com/centrifuge/connectors/pull/16#issue-1315036467?

AStox:
Agreed 👍

PR #30, opened 2022-08-12
fix fuzzed tests hitting global reject limit

AStox:
Some fuzzed tests were hitting the global reject limit, caused by a `vm.assume` call being too non-specific. This PR re-writes those tests to avoid the need for the offending `vm.assume` call.

### DIFF
--- a/test/Connector.t.sol
+++ b/test/Connector.t.sol
@@ -81,8 +81,9 @@ contract ConnectorTest is Test {
         homeConnector.addTranche(poolId, trancheId, tokenName, tokenSymbol);
     }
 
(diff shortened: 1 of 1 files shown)


AStox:
So far this PR seems to have successfully fixed tests that were breaking due to the global fuzzing rejects threshold, but now I'm wondering if we shouldn't also increase that threshold for our PR tests, since we've increased the fuzz runs a significant amount as well.

https://book.getfoundry.sh/reference/config#fuzz_max_global_rejects

Default is set to 65536, but each test is run 50,000 times now. Maybe an increase to 200,000? What do we think? @offerijns @ilinzweilin

offerijns:
Makes sense to me to raise that as well, @AStox

offerijns (approved):

PR #31, opened 2022-11-22
Fix AddTranche

NunoAlexandre:
Fix `AddTranche` by:

1. Splitting the act of _adding_ the tranche and _deploying_ it into two separate functions
2. Support the new `AddTranche` message format, expecting a `tokenName` of 128 bytes instead of 32

On 2., we are breaking this into two iterations:

a. We just parse the first 32 bytes out of the 128 bytes allocated for the `tokenName` and proceed with reading `tokenSymbol` after said `128` bytes; Meaning we will get a truncated but working version of `tokenName`

b. Support the full 128 bytes for `tokenName`; this will probably require (de)encoding 4 blocks of 32 bytes since we don't find primitive support for bytearrays longer than 32 bytes.


NunoAlexandre:
@offerijns the `testAddingSingleTrancheWorks`, amongst other tests, is failing and I could make some steps forward but I need help here.

The first issue I found was that, since the `Connector::addTranche` function no longer deploys the tranche, the `token_` obtained when we were then calling `bridgedConnector.tranches` would be null; calling `bridgedConnector.deployTranche(poolId, trancheId);` beforehand did the trick.

However, since we call `addTranche` in `homeConnector` BUT `deployTranche` in `bridgeConnector`, the latter doesn't know about the token name and symbol, so the tests on `RestrictedTokenLike` all fail because those attributes are empty.

Are we adding a new Connectors message `DeployTranche`? That would allow us to add `deployTranche` to `homeConnector` which would ensure that both `addTranche` and `deployTranche` follow the same route and therefore the same state.

Let me have your thoughts on this as I may be missing something as well.

NunoAlexandre:
@AStox @offerijns if we go with iteration 1. first, this PR is ready :)

offerijns (dismissed):
I think it would be good to also get @AStox's approval on this, but from my perspective this looks ready to be merged! Awesome first Solidity PR @NunoAlexandre 🔥

src/Connector.sol
```diff
@@ -26,6 +26,8 @@ contract CentrifugeConnector is Test {
         address token;
         uint256 latestPrice; // [ray]
         uint256 lastPriceUpdate;
+        string tokenName;
```
offerijns, line 29: One thought: can you add a todo here like
```suggestion
        // TODO: the token name & symbol need to be stored because of the separation between adding and deploying tranches.
        // This leads to duplicate storage (also in the ERC20 contract), ideally we should refactor this somehow
        string tokenName;
```

NunoAlexandre:
> I think it would be good to also get @AStox's approval on this, but from my perspective this looks ready to be merged! Awesome first Solidity PR @NunoAlexandre 🔥

Awesome :) will wait for @AStox. Thank you 🚀

NunoAlexandre:
@AStox would you have time to have a look? Totally understand if you don't have the time now 🙏

AStox (changes requested):
Looks great Nuno! Good job! While there's nothing wrong with your PR, I think we should remove the `AddTranche()` split logic and just merge the `Support the new AddTranche message format` changes.

I think splitting add tranche should stay on an experimental branch for now, while we figure out what can be done.

AStox (approved):
Nevermind, just learned that `deployTranche()` won't need to be called via XCM, so the gas restrictions are solved!

🚢 🚢 🚢

PR #32, opened 2023-01-30
Fix `UpdateMember` encoding & decoding

NunoAlexandre:
When testing the latest contract revision in `add-withdaw-deposit`, `UpdateMember` failed with a parsing error.
Here we fix all of these issues and improve the tests to verify the correctness of those changes. On the cent-chain side we tweak the encoding of the `UpdateMember.validUntil` to little endian so that it is parsed correctly on this side of the system.

### DIFF
--- a/src/Connector.sol
+++ b/src/Connector.sol
@@ -141,7 +141,7 @@ contract CentrifugeConnector {
         uint64 poolId,
         bytes16 trancheId,
         address user,
-        uint256 validUntil
+        uint64 validUntil
     ) public onlyRouter {
         Tranche storage tranche = tranches[poolId][trancheId];
         require(tranche.latestPrice > 0, "CentrifugeConnector/invalid-pool-or-tranche");
--- a/src/Messages.sol
+++ b/src/Messages.sol
@@ -117,16 +117,41 @@ library ConnectorMessages {
         validUntil = uint256(_msg.index(45, 32));
     }
 
-    function parseUpdateMemberDebug(bytes29 _msg) internal pure returns (uint64 poolId, bytes16 trancheId) {
+    function parseUpdateMemberDebug(bytes29 _msg) internal pure returns (uint64 poolId, bytes16 trancheId, address user, uint64 validUntil) {
         poolId = uint64(_msg.indexUint(1, 8));
         trancheId = bytes16(_msg.index(9, 16));
-//        user = address(bytes20(_msg.index(25, 20)));
-//        // TODO: skip 12 padded zeroes from address
-//        validUntil = uint256(_msg.index(45, 32));
+        user = address(bytes20(_msg.index(25, 20)));
+        validUntil = uint64(_msg.indexUint(57, 8));
     }
 
-    function parseTrancheId(bytes29 _msg) internal pure returns (bytes16 trancheId) {
-        trancheId = bytes16(_msg.index(0, 16));
+    // Convert an hexadecimal character to their value
+    function fromHexChar(uint8 c) internal pure returns (uint8) {
+        if (bytes1(c) >= bytes1("0") && bytes1(c) <= bytes1("9")) {
+            return c - uint8(bytes1("0"));
+        }
+        if (bytes1(c) >= bytes1("a") && bytes1(c) <= bytes1("f")) {
+            return 10 + c - uint8(bytes1("a"));
+        }
+        if (bytes1(c) >= bytes1("A") && bytes1(c) <= bytes1("F")) {
+            return 10 + c - uint8(bytes1("A"));
+        }
+        revert("Failed to encode hex char");
+    }
+
+    // Convert an hexadecimal string to raw bytes
+    function fromHex(string memory s) internal pure returns (bytes memory) {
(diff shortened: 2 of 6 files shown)


src/Connector.sol
```diff
@@ -141,7 +141,7 @@ contract CentrifugeConnector {
         uint64 poolId,
         bytes16 trancheId,
         address user,
-        uint256 validUntil
+        uint64 validUntil
```
NunoAlexandre, line 144: This was wrong and often `validUntil` was named `amount`, probably from a copy-paste from `transfer` I guess

src/Messages.sol
```diff
@@ -96,25 +96,55 @@ library ConnectorMessages {
      * 0: call type (uint8 = 1 byte)
      * 1-8: poolId (uint64 = 8 bytes)
      * 9-25: trancheId (16 bytes)
-     * 26-46: user (Ethereum address, 20 bytes)
-     * 47-78: validUntil (uint256 = 32 bytes)
+     * 25-45: user (Ethereum address, 20 bytes - Skip 12 bytes from 32-byte addresses)
+     * 57-65: validUntil (uint64 = 8 bytes)
      * 
      * TODO: use bytes32 for user (for non-EVM compatibility)
```
NunoAlexandre, line 102: @offerijns I started working on this but we would need to also update `Transfer` and the storage items storing `address` now. I prefer to do that in a follow up PR unless if you think we shouldn't do this now?

src/Messages.sol
```diff
@@ -96,25 +96,55 @@ library ConnectorMessages {
...
      */
-    function formatUpdateMember(uint64 poolId, bytes16 trancheId, address user, uint256 validUntil) internal pure returns (bytes memory) {
-        return abi.encodePacked(uint8(Call.UpdateMember), poolId, trancheId, user, validUntil);
+    function formatUpdateMember(uint64 poolId, bytes16 trancheId, address user, uint64 validUntil) internal pure returns (bytes memory) {
+        return abi.encodePacked(uint8(Call.UpdateMember), poolId, trancheId, user, bytes(hex"000000000000000000000000"), validUntil);
```
NunoAlexandre, line 105: Since the decoding parses the first 20 bytes of `user` and skips the following 12, here we need to append 12 padding zeros to make it right.

test/Messages.t.sol
```diff
@@ -91,41 +91,41 @@ contract MessagesTest is Test {
 
     function testUpdateMemberEncoding() public {
         assertEq(
-            ConnectorMessages.formatUpdateMember(5, toBytes16(fromHex("010000000000000003")), 0x225ef95fa90f4F7938A5b34234d14768cB4263dd, 1657870537), 
-            fromHex("04000000000000000500000000000000000000000000000009225ef95fa90f4f7938a5b34234d14768cb4263dd0000000000000000000000000000000000000000000000000000000062d118c9")
-            );
+            ConnectorMessages.formatUpdateMember(2, bytes16(hex"811acd5b3f17c06841c7e41e9e04cb1b"), 0x1231231231231231231231231231231231231231, 1706260138),
```
NunoAlexandre, line 94: An interesting and frustration finding was that `toBytes16(fromHex("811acd5b3f17c06841c7e41e9e04cb1b"))` was retuning me an empty `bytes16` 🤷‍♂️

NunoAlexandre (commented):
Some info for the reviewers

src/Connector.sol
offerijns: Good find!

src/Messages.sol
offerijns: I think that's totally fine in a separate PR!

offerijns (approved):
Looks good to me!

src/Messages.sol
NunoAlexandre: @offerijns done 👍

src/Messages.sol
AStox: Yah separating issues into separate PRs is a good idea. I'll note here so I don't forget though, it looks like we're double-reading a byte here. Tranche ID is 9-25, while user is 25-45.

test/Messages.t.sol
AStox: Odd... does `fromHex()` expect a prepended `0x` maybe?

AStox: Doesn't really look like it based on how we use it elsewhere actually.

src/Messages.sol
NunoAlexandre: @AStox I first thought this as well but it doesn't seem to be the case:

``` solidity
  poolId = uint64(_msg.indexUint(1, 8));
  trancheId = bytes16(_msg.index(9, 16));
  user = address(bytes20(_msg.index(25, 20)));
  validUntil = uint64(_msg.indexUint(57, 8));
```

So, `poolId` would be `1 + 8 = 9`, and then we would be reading `trancheId` from byte at index `9`, which would be another example of double-reading.

The docs were wrong indeed tho, as it's `9-24`. Thanks for the catch!

PR #35, opened 2023-01-31
Feat/domain update

AStox:


.gitmodules
```diff
@@ -7,4 +7,4 @@
 [submodule "lib/memview-sol"]
 	path = lib/memview-sol
 	url = https://github.com/summa-tx/memview-sol
-	branch = main
+	branch = v2.1.1
```
NunoAlexandre, line 10: What's the reason to change this? I had issues with setting the branch to `v2.1.1` last week and could only get it working with `main`

src/Messages.sol
```diff
@@ -159,8 +159,8 @@ library ConnectorMessages {
      * 47-78: amount (uint256 = 32 bytes)
      * 
      */
-    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount) internal pure returns (bytes memory) {
-        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, amount);
+    function formatTransfer(uint32 destinationDomain, uint64 poolId, bytes16 trancheId, address user, uint256 amount) internal pure returns (bytes memory) {
```
NunoAlexandre, line 162: Let's move this argument to the end so that it matches with the encoding order 

```suggestion
    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount, uint32 destinationDomain) internal pure returns (bytes memory) {
```

test/mock/MockHomeConnector.sol
```diff
@@ -53,8 +53,8 @@ contract MockHomeConnector is Test {
         router.handle(_message);
     }
 
-    function transfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount) public  {
-        bytes memory _message = ConnectorMessages.formatTransfer(poolId, trancheId, user, amount);
+    function transfer(uint32 destinationDomain, uint64 poolId, bytes16 trancheId, address user, uint256 amount) public  {
+        bytes memory _message = ConnectorMessages.formatTransfer(destinationDomain, poolId, trancheId, user, amount);
```
NunoAlexandre, line 57:
```suggestion
    function transfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount, uint32 destinationDomain) public  {
        bytes memory _message = ConnectorMessages.formatTransfer(poolId, trancheId, user, amount, destinationDomain);
```

src/Connector.sol
```diff
@@ -167,16 +159,14 @@ contract CentrifugeConnector {
         bytes16 trancheId,
         address user,
         uint256 amount,
-        string memory domainName
+        uint32 destinationDomain
```
NunoAlexandre, line 162: Are we ok with it being an arbirtrary `uint32` value? I thought we would also have an enum and restrict a `domain` argument to be one of that enum's variants?

NunoAlexandre: cc @offerijns

.gitmodules
AStox: Oh really? Okay yah this is the stupidest thing. Sometimes it works for me on main, sometimes I need to specify the version. I had explained it to the foundry team at the beginning of January and they said it was a bug and were going to fix it. For now I'll remove this change, but hopefully we can get on the same page with our lib/ soon lol

src/Connector.sol
offerijns: > I thought we would also have an enum and restrict a domain argument to be one of that enum's variants?

Yeah this is what should happen

src/Messages.sol
```diff
@@ -159,8 +159,8 @@ library ConnectorMessages {
      * 47-78: amount (uint256 = 32 bytes)
      * 
      */
-    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount) internal pure returns (bytes memory) {
-        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, amount);
+    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount, uint32 destinationDomain) internal pure returns (bytes memory) {
+        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, amount, destinationDomain);
     }
 
     function isTransfer(bytes29 _msg) internal pure returns (bool) {
```
NunoAlexandre, line 166: Can't leave a comment there but we are forgetting to parse the domain from a `Transfer` message. However, I am not sure I understand how useful it is to have that domain at the Connector contract level? 🤔

NunoAlexandre (approved):

NunoAlexandre (changes requested):
Sorry, didn't mean to approve. See comments above 🙏

src/Messages.sol
AStox: Yah, I'm not sure I understand why we're sending the domain to the connector. Seems like it should only be the router that cares about that domain for Cent -> EVM messages, right?

AStox: @NunoAlexandre I propose we leave it out until we can figure out a reason to have it there.

src/Connector.sol
```diff
@@ -7,11 +7,15 @@ import { RestrictedTokenLike } from "./token/restricted.sol";
 import { MemberlistLike } from "./token/memberlist.sol";
 
 interface RouterLike {
-    function sendMessage(uint32 destinationDomain, uint64 poolId, bytes16 trancheId, uint256 amount, address user) external;
+    function sendMessage(uint64 poolId, bytes16 trancheId, uint256 amount, address user) external;
 }
 
 contract CentrifugeConnector {
 
+    enum Domains {
```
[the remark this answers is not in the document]
AStox: Not sure I understand. Your enums resolve to `0x000100000000000000`, for example?

src/Messages.sol
NunoAlexandre: @AStox let's get @offerijns pov here since he had suggested adding the field to the message.

src/Connector.sol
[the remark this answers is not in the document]
NunoAlexandre: > Not sure I understand. Your enums resolve to 0x000100000000000000, for example?


Each Domain variant, i.e, `Domain::EVM(_)` or `Domain::Parachain(_)` will encode different. See the individual tests linked above for each variant and with different values. The `Moonbeam` domain encodes to `0x0100`, the ethereum mainnet encodes to `0x000100000000000000`, etc. On the cent-chain we currently don't handle `Centrifuge` but we should add it in order to be able to decode `Transfer` messages with `domain` set `Centrifuge` (again, if we keep that field in that message)

src/Messages.sol
offerijns: We need it for EVM > EVM transfers in the future, since they will go through Cent Chain and thus will need to know the target domain. While we don't need to implement it today, it would be a shame if ew have to change the Transfer message format later.

AStox: @offerijns what about the `parseTransfer()` function. No reason to be sending the domain to the connector for an incoming message, right?

NunoAlexandre: Makes sense @offerijns!

And that also makes sense @AStox

offerijns: Do you want to have two different `Transfer` messages, one that goes to Cent Chain and one that goes to Connectors? Or do I misunderstand?

NunoAlexandre: @offerijns i think Adam means that there is no point currently in actually parsing the domain field from the Transfer message. I think this is fine but it's also fine to already have that code in place and tested. Wdyt?

AStox: Yah I thought Nuno's original question here was why aren't we parsing the domain from the `Transfer` message in `parseTransfer()`. Maybe I misunderstood, but I think domain *does* belong in the Transfer message, but when parsing it for the Connector, there's no point in also parsing and sending the domain

src/Connector.sol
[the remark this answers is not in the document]
AStox: @offerijns How could this be implemented in Solidity? I don't think we have support for enum variants

src/Messages.sol
offerijns: True, we don't need to parse it now, but we should make sure to already define the length of the domain param. Otherwise if in the future this changes, we'll have to change the parsing of the arguments after it and that would not be a backwards compatible change.

src/Connector.sol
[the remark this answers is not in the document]
offerijns: I don't know, that needs some research! Simplest solution would be manually encoding it the same way as cnet chain. So `bytes(uint8(0) )+ (uint64(evmChainId))` for an EVM chain id basically and `bytes(uint8(1) )+ (uint64(centchainIarachainId))` for Cent Chain.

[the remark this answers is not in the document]
AStox: Then Alina's `DomainLookup` that we removed might be the best approach to solve this. Our reasoning for removing it was that we didn't want the extra SLOAD, but I think in order to use an enum pattern that resolves to real data (instead of just integers), it will always involve an SLOAD.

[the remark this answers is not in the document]
offerijns: I think the function interface can just be
```solidity
enum Domain {EVM, Parachain}

function transfer(
    uint64 poolId,
    bytes16 trancheId,
    address user,
    uint256 amount,
    Domain destinationDomain,
    uint destinationDomainId
) public {}
```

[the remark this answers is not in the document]
AStox: Alright just pushed an update for this I think we'll all be happy with. Still need to work on the encoding though. @NunoAlexandre maybe we can work on this together tomorrow. I'm not really sure how you're encoding the chainIDs to match whats show in your tests.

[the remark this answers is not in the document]
NunoAlexandre: @AStox sounds good! For the example test you wrote for Avalanche, I see that the Avalanche chain Id encodes as little endian on Solidity, so we need to do that on the cent-chain as well (we had to do that for a couple of other uint values already). Notice how on the rust side it encodes to `6aa80` while on the solidity side it is `a86a` (reverse byte order)

src/Connector.sol
```diff
@@ -81,15 +83,11 @@ contract CentrifugeConnector {
...
     }
 
-    function file(bytes32 name, string memory domainName, uint32 domainId) public auth  {
-        if(name == "domain") {
-           domainLookup[keccak256(bytes(domainName))] = domainId;
-           emit File(name, domainName);
-        } else { revert ("unknown name");}
-        
+    // --- Internal ---
+    function getDomain(Domain domain, uint domainId) public pure returns (bytes32) {
```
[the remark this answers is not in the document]
AStox: Agreed!

src/Connector.sol
```diff
@@ -81,15 +83,11 @@ contract CentrifugeConnector {
...
+        return bytes32(uint64(domain) + domainId);
```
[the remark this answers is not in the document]
AStox: I like it! Gonna implement your suggestion

test/Connector.t.sol
```diff
@@ -176,6 +176,11 @@ contract ConnectorTest is Test {
         bridgedConnector.updateTokenPrice(poolId, trancheId, price);
      }
 
+    function testGetDomainAvalanche() public {
```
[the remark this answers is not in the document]
AStox: Yah now that I've implemented your suggestion above I think we can get rid of the constant

src/Connector.sol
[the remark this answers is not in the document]
offerijns: I just realised, to your comment earlier @NunoAlexandre that Moonbeam actually is the EVM itself, so it would be identified by the EVM domain with the EVM chain ID. And given that we're quite unlikely to implement Connectors in anything other than Solidity / targeting the EVM for quite a while. Maybe we should just simplify this to

```
enum Domain { EVM, Centrifuge }
```

And make `domainId` optional. That way if the target is an EVM, then we use chain IDs, and if the target is Centrifuge, there's no additional ID required. In the far future when we want to implement non-EVM targets, we can re-add a `Parachain` domain, but it still makes sense to have `Centrifuge` separate as it encodes more efficiently.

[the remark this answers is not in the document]
NunoAlexandre: @offerijns cool, this starts making more conceptual sense to me. However, how do we map this logic into the cent-chain, specially when it comes to special cases like Moonbeam? I guess we would specify "Moonbeam" by specifying its EVM Chain ID and we would internally look up the Router associated with that domain?

src/Connector.sol
```diff
@@ -5,9 +5,10 @@ pragma abicoder v2;
 import { RestrictedTokenFactoryLike, MemberlistFactoryLike } from "./token/factory.sol";
 import { RestrictedTokenLike } from "./token/restricted.sol";
 import { MemberlistLike } from "./token/memberlist.sol";
+import { ConnectorMessages } from "src/messages.sol";
```
offerijns, line 8: I guess this is why the tests are failing:
```suggestion
import { ConnectorMessages } from "src/Messages.sol";
```

src/Connector.sol
```diff
@@ -168,16 +159,14 @@ contract CentrifugeConnector {
         bytes16 trancheId,
         address user,
         uint256 amount,
-        string memory domainName
+        ConnectorMessages.Domain domain
```
offerijns, line 162: I don't really like having the Connector smart contract depend on the Messages library. I think it would be good if we can make it so that any changes to messages don't impact the Connector. But we can keep this for v2, when we refactor it, likely with the introduction of a Gateway contract inbetween.

offerijns (approved):
LGTM

src/Messages.sol
```diff
@@ -174,4 +176,12 @@ library ConnectorMessages {
         amount = uint256(_msg.index(45, 32));
     }
 
+    function formatDomain(Domain domain) public pure returns (bytes9) {
+        return bytes9(byte(uint8(domain)));
+    }
+
+    function formatDomain(Domain domain, uint64 domainId) public pure returns (bytes memory) {
+        return abi.encodePacked(uint8(domain), domainId);
+    }
```
NunoAlexandre, line 185: We should have both `formatDomain` functions returning `bytes9` to ensure - at the type level - that they both generate a 9-long byte encoded representation of a domain.

```suggestion
    function formatDomain(Domain domain, uint64 domainId) public pure returns (bytes9) {
        return bytes9(abi.encodePacked(uint8(domain), domainId).ref(0).index(0, 9));
    }
```

test/Messages.t.sol
```diff
@@ -171,6 +171,22 @@ contract MessagesTest is Test {
         assertEq(uint(decodedPrice), uint(price));
     }
 
+    function testFormatDomainCentrifuge() public {
+        assertEq(ConnectorMessages.formatDomain(ConnectorMessages.Domain.Centrifuge), hex"00");
```
NunoAlexandre, line 175: Let's make this 9 bytes as well to be explicit:

```suggestion
        assertEq(ConnectorMessages.formatDomain(ConnectorMessages.Domain.Centrifuge), hex"000000000000000000");
```

src/Messages.sol
```diff
@@ -159,8 +161,8 @@ library ConnectorMessages {
      * 47-78: amount (uint256 = 32 bytes)
      * 
      */
-    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount) internal pure returns (bytes memory) {
-        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, amount);
+    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount, uint32 destinationDomain) internal pure returns (bytes memory) {
```
NunoAlexandre, line 164: This is not correct since `domain` is not `uint32`. Either we receive `bytes9` here and expect the caller of `formatTransfer` to first call `formatDomain` first, or we overload `formatTransfer for 1) `(Domain domain) and 2) Domain domain, uint64 domainId)` and call the right `formatDomain` internally, respectively.

src/Connector.sol
```diff
@@ -35,14 +36,12 @@ contract CentrifugeConnector {
...
     mapping(address => uint256) public wards;
-    mapping(bytes32 => uint32) public domainLookup;
-
 
     // --- Events ---
     event Rely(address indexed user);
     event Deny(address indexed user);
     event File(bytes32 indexed what, address data);
-    event File(bytes32 indexed what, string data);
+    event File(bytes32 indexed what, string data, uint32 data2);
```
NunoAlexandre, line 44: Do we still need this change?

NunoAlexandre (changes requested):
We have small things to fix first. And I thought we had tests for the `transfer` message encoding? But I don't see them anymore? For me it's alright, I can add them once I make the rust side match this new implementation of the message and the domain type and then ensure that both parts can encode/decode the same message 👍

src/Connector.sol
NunoAlexandre: That's a good point Jeroen. I suggested to @AStox to move the formatting of the domain to `Messages` to have that function alongside all the other encoding/decoding functions. Let's tackle this when we go for v2 👍

src/Messages.sol
AStox: I think changing it to bytes9 is the cleanest solution

src/Connector.sol
AStox: Nope! Removed.

AStox:
Okay @NunoAlexandre should be all good. I added in the transfer encoding/decoding test but just using dummy data, so we can replace it later with some real payloads from cent chain.

PR #36, opened 2023-02-01
Align tests with cent chain

NunoAlexandre:


test/Messages.t.sol
```diff
@@ -82,26 +82,31 @@ contract MessagesTest is Test {
...
             ConnectorMessages.formatUpdateMember(2, bytes16(hex"811acd5b3f17c06841c7e41e9e04cb1b"), 0x1231231231231231231231231231231231231231, 1706260138),
             hex"040000000000000002811acd5b3f17c06841c7e41e9e04cb1b12312312312312312312312312312312312312310000000000000000000000000000000065b376aa"
         );
     }
 
+    // We use an UpdateMember encoded message generated in the cent-chain to
+    // verify we handle the 32 to 20 bytes address compatibility as expected.
     function testUpdateMemberDecoding() public {
-        (uint64 decodedPoolId, bytes16 decodedTrancheId, address decodedUser, uint64 decodedValidUntil) = ConnectorMessages.parseUpdateMember(fromHex("040000000000000002811acd5b3f17c06841c7e41e9e04cb1b12312312312312312312312312312312312312312312312312312312312312310000000065B376AA").ref(0));
+        (uint64 decodedPoolId, bytes16 decodedTrancheId, address decodedUser, uint64 decodedValidUntil) = ConnectorMessages.parseUpdateMember(fromHex("040000000000000002811acd5b3f17c06841c7e41e9e04cb1b12312312312312312312312312312312312312312312312312312312312312310000000065b376aa").ref(0));
```
NunoAlexandre, line 109: @offerijns note how this test uses a `UpdateMember` message generated by the cent-chain which encodes a 32-byte address. The solidity side parses the message well, which includes parsing that 32-byte address as a 20-byte one.

Ideally, both solidity and the cent-chain sides would generate the exact same `UpdateMember` message but on solidity side it will be different since the address will be encoded as a `20 bytes` + `12 zeros`.

We do have a `To Do` to make `address user` be `bytes32 user` instead and for me this is a good time to implement this. Do you agree or am I overseeing something?

NunoAlexandre:
@AStox this is ready. we just missed one test to be aligned with the cent-chain side in the end

AStox (approved):

PR #39, opened 2023-02-07
Fine-tune transfers encoding, decoding, and tests

NunoAlexandre:
Here we align the encoding/decoding tests of the `Transfer` message using cent-chain generated messages. We fix some issues with the current encoding and decoding functions and extend the tests to cover more use cases.

### DIFF
--- a/src/Messages.sol
+++ b/src/Messages.sol
@@ -157,17 +157,11 @@ library ConnectorMessages {
      * 0: call type (uint8 = 1 byte)
      * 1-8: poolId (uint64 = 8 bytes)
      * 9-24: trancheId (16 bytes)
-     * 25-56: user (Ethereum address, 20 bytes)
+     * 25-56: user (Ethereum address, 20 bytes - Skip last 12 bytes for 32-byte address compatibility)
      * 57-72: amount (uint128 = 16 bytes)
      * 73-82: domain (Domain = 9 bytes)
      */
     function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint128 amount, bytes9 destinationDomain) internal pure returns (bytes memory) {
-        // 1 byte - call type
-        // 8 bytes - poolId
-        // 16 bytes - TrancheId
-        // 32 bytes - Address
-        // 16 bytes - Amount
-        // 9 bytes - Domain
         return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, bytes(hex"000000000000000000000000"), amount, destinationDomain);
     }
 
@@ -176,18 +170,11 @@ library ConnectorMessages {
     }
 
     function parseTransfer(bytes29 _msg) internal pure returns (uint64 poolId, bytes16 trancheId, address user, uint128 amount, bytes9 encodedDomain) {
-        // 8 bytes - Pool id
-        // 16 bytes - tranche Id
-
-        // 9 bytes - domain
-        // 32 bytes - address
-        // 16 bytes - Amount (????)
-
         poolId = uint64(_msg.indexUint(1, 8));
         trancheId = bytes16(_msg.index(9, 16));
         user = address(bytes20(_msg.index(25, 20)));
-        amount = uint128(_msg.indexUint(45, 16));
-        encodedDomain = bytes9(_msg.index(45, 9));
+        amount = uint128(_msg.indexUint(57, 16));
+        encodedDomain = bytes9(_msg.index(73, 9));
     }
 
     function formatDomain(Domain domain) public pure returns (bytes9) {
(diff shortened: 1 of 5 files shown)


src/Messages.sol
```diff
@@ -156,24 +156,25 @@ library ConnectorMessages {
...
-     * 
+     * 9-24: trancheId (16 bytes)
+     * 25-56: user (Ethereum address, 20 bytes - Skip last 12 bytes for 32-byte address compatibility)
+     * 57-72: amount (uint128 = 16 bytes)
+     * 73-82: domain (Domain = 9 bytes)
      */
-    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount, bytes9 destinationDomain) internal pure returns (bytes memory) {
-        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, amount, destinationDomain);
+    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint128 amount, bytes9 destinationDomain) internal pure returns (bytes memory) {
+        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, bytes(hex"000000000000000000000000"), amount, destinationDomain);
```
NunoAlexandre, line 165: We needed to append 12 zeros here to make the address encode to a total of 32 bytes

src/Messages.sol
```diff
@@ -156,24 +156,25 @@ library ConnectorMessages {
...
-     * 47-78: amount (uint256 = 32 bytes)
-     * 
+     * 9-24: trancheId (16 bytes)
+     * 25-56: user (Ethereum address, 20 bytes - Skip last 12 bytes for 32-byte address compatibility)
+     * 57-72: amount (uint128 = 16 bytes)
+     * 73-82: domain (Domain = 9 bytes)
      */
-    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount, bytes9 destinationDomain) internal pure returns (bytes memory) {
-        return abi.encodePacked(uint8(Call.Transfer), poolId, trancheId, user, amount, destinationDomain);
+    function formatTransfer(uint64 poolId, bytes16 trancheId, address user, uint128 amount, bytes9 destinationDomain) internal pure returns (bytes memory) {
```
NunoAlexandre, line 164: We needed to encode amount as `uint128`

NunoAlexandre (commented):
Some notes

src/Messages.sol
AStox: I'm a little unsure of using a `u128` for `amount` since balances are generally `u256` in solidity, but I don't know if it really matters. If we run into issues later we can always change it.

src/Messages.sol
```diff
@@ -156,24 +156,25 @@ library ConnectorMessages {
...
      * 0: call type (uint8 = 1 byte)
      * 1-8: poolId (uint64 = 8 bytes)
-     * 9-25: trancheId (16 bytes)
-     * 26-46: user (Ethereum address, 20 bytes)
-     * 47-78: amount (uint256 = 32 bytes)
-     * 
+     * 9-24: trancheId (16 bytes)
+     * 25-56: user (Ethereum address, 20 bytes - Skip last 12 bytes for 32-byte address compatibility)
+     * 57-72: amount (uint128 = 16 bytes)
+     * 73-82: domain (Domain = 9 bytes)
```
AStox, line 162: Domain bytes should be 73-81 no? Since 72 + 9 = 81

test/Messages.t.sol
```diff
@@ -171,30 +171,47 @@ contract MessagesTest is Test {
...
-        (uint64 decodedPoolId, bytes16 decodedTrancheId, address decodedUser, uint256 decodedAmount) = ConnectorMessages.parseTransfer(_message.ref(0));
+    function testTransferEquivalence(uint64 poolId, bytes16 trancheId, address user, uint128 amount, uint64 destinationChainId) public {
+        bytes9 inputEncodedDomain = ConnectorMessages.formatDomain(ConnectorMessages.Domain.EVM, destinationChainId);
+        bytes memory _message = ConnectorMessages.formatTransfer(poolId, trancheId, user, amount, inputEncodedDomain);
+        (uint64 decodedPoolId, bytes16 decodedTrancheId, address decodedUser, uint256 decodedAmount, bytes9 encodedDomain) = ConnectorMessages.parseTransfer(_message.ref(0));
         assertEq(uint(decodedPoolId), uint(poolId));
         assertEq(decodedTrancheId, trancheId);
         assertEq(decodedUser, user);
         assertEq(decodedAmount, amount);
+        assertEq(encodedDomain, inputEncodedDomain);
```
AStox, line 214: Woops! I forgot to test the most important part! Good catch 👍

AStox (approved):

src/Messages.sol
NunoAlexandre: I remember talking about this with @offerijns and the conclusion was that u128 was enough for all operations we need. u128 is the type used in the cent-chain so that type should be big enough to cover tranche transfers in the EVM as well 👍

PR #40, opened 2023-02-09
Set up invariant test

offerijns:


test/accounts/PoolManager.sol
```diff
@@ -0,0 +1,40 @@
...
+  bytes16[] public allTranches;
+  mapping(bytes16 => uint64) public trancheIdToPoolId;
+
+  constructor(MockHomeConnector connector_) public {
+    connector = connector_;
+  }
+
+  function addPool(uint64 poolId) public {
+    connector.addPool(poolId);
+
```
NunoAlexandre, line 22:
```suggestion
```

test/accounts/PoolManager.sol
```diff
@@ -0,0 +1,40 @@
...
+  function addPool(uint64 poolId) public {
+    connector.addPool(poolId);
+    
+    allPools.push(poolId);
+  }
+
+  // uint8 so the fuzzer more often finds valid array indices
+  function addTranche(uint8 poolIndex, bytes16 trancheId, string memory tokenName, string memory tokenSymbol, uint128 price) public {
+    uint64 poolId = allPools[poolIndex];
+    vm.assume(poolId > 0);
```
NunoAlexandre, line 29: Why do we need this `assume`?

NunoAlexandre (dismissed):
This is nice 👌

test/accounts/PoolManager.sol
offerijns: The fuzzer will generate random uint8 `poolIndex` values, and there's no guarantee that these actually exist. This basically checks that.

NunoAlexandre (approved):

PR #45, opened 2023-02-16
Run `forge fmt`

NunoAlexandre:
I noticed the code is all over the place in some parts so running `forge fmt` to get it all fmt'ed. We should start running `forge fmt` before committing going further 🌟

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -6,7 +6,7 @@ env:
   FOUNDRY_PROFILE: ci
 
 jobs:
-  check:
+  test:
     strategy:
       fail-fast: true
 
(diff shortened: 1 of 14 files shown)


.github/workflows/pull-request.yml
```diff
@@ -34,4 +34,28 @@ jobs:
         id: test
 
         env:
-          FOUNDRY_PROFILE: push_to_main
\ No newline at end of file
+          FOUNDRY_PROFILE: push_to_main
```
offerijns, line 37:
```suggestion
          FOUNDRY_PROFILE: pull_request
```

offerijns: Not sure why but looks like this was somehow set incorrectly in the past

offerijns (approved):
LGTM!

PR #46, opened 2023-02-22
feat: send Transfer message

NunoAlexandre:


src/Connector.sol
```diff
@@ -134,14 +134,14 @@ contract CentrifugeConnector {
         memberlist.updateMember(user, validUntil);
     }
 
-    function handleTransfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount) public onlyRouter {
+    function handleTransfer(uint64 poolId, bytes16 trancheId, address user, uint128 amount) public onlyRouter {
```
mustermeiszer, line 137: I thought we agreed on having `u256` as the amount or is this not related to the message part?

NunoAlexandre: We didn't agree on having `u256` tho, we agreed on `u128` since that's the type that is supported on Centrifuge chain, the core product. We can always coerce `uint128` to `u256` safely (for Ethereum integrations) but we can't coerce `u256` to `u128`. The only way around this would be by changing the `Balance` type in Centrifuge to `u256`.

What are your thoughts on this?

src/routers/xcm/Router.sol
```diff
@@ -44,6 +69,11 @@ contract ConnectorXCMRouter {
         _;
     }
 
+    // todo(nuno): add auth modifier
+    function update(XcmWeightInfo memory xcmWeightInfo_) external {
+        xcmWeightInfo = xcmWeightInfo_;
+    }
```
NunoAlexandre, line 75: @offerijns @AStox I don't know what's the Solidity way of doing this. The `file` pattern doesn't really seem applicable here because we are not handling random `bytes` data as input.

src/Connector.sol
NunoAlexandre, line 137: @mustermeiszer @offerijns I also realised that this API won't allow to transfer tokens to an actual centrifuge address, since that currently the API expects an Ethereum `address` that we suffix with 12 zeros for Centrifuge compatibility.

src/routers/xcm/Router.sol
offerijns: I think we can still use the `file` pattern, just
```solidity
 function file(bytes32 what, XcmWeightInfo memory xcmWeightInfo_) external auth {
        if (what == "weight-info") xcmWeightInfo = xcmWeightInfo_;
        else revert("XCMRouter/file-unrecognized-param");
        emit File(what, data);
 }
```

src/Connector.sol
offerijns: `handleTransfer` is only for incoming transfers though, right? Which always should go to an Ethereum `address`

offerijns: `transfer` definitely should accept a `bytes32` based address though

src/routers/xcm/Router.sol
NunoAlexandre: I don't get that tho. Why having a "what" if it clearly only works to set the XcmWeightInfo?

src/Connector.sol
NunoAlexandre: Yeah sorry, left the comment in the wrong place 👍🏻

src/routers/xcm/Router.sol
AStox: This is most likely going to be called via multi-sig, and I'm not sure how well 3rd party UIs like gnosis safe or etherscan will parse structs. In Tinlake, for things like rate groups, we pass in all the parameters individually and instantiate a new struct.

```solidity
function file(bytes32 what, uint64 buyExecutionWeightLimit, uint64 transactWeightAtMost, uint256 feeAmount) external auth { 
    if (what == "weight-info") {
        xcmWeightInfo = xcmWeightInfo(buyExecutionWeightLimit, transactWeightAtMost, feeAmount);
    } else {
        revert("XCMRouter/file-unrecognized-param"); 
    }
    emit File(what, data); 
}
```

src/routers/xcm/Router.sol
```diff
@@ -61,15 +91,124 @@ contract ConnectorXCMRouter {
...
+            buyExecutionWeightLimit
+        );
+    }
+
+    function centrifuge_handle_call(bytes memory message) internal view returns (bytes memory) {
+        return abi.encodePacked(
+            // The call index; first byte is the pallet, the second is the extrinsic
+            centrifugeChainHandleCallIndex,
+            // We need to specify the length of the message in the scale-encoding format
+            // A transfer message has 82 bytes which encodes to 4901 in Scale :shrug:
```
AStox, line 165: lol 🤷‍♂️

src/Connector.sol
mustermeiszer: Are, then I was wrong. No that is perfectly fine. Had just another memory. Just need to remember to adapt the spec ^^

NunoAlexandre: Cool :)

src/routers/xcm/Router.sol
NunoAlexandre: Good insight, @AStox. thanks! I guess the only purpose of `what` here is to have that identifying the `File` event? But then again, in your example you file `data` where that doesn't apply in this case.

I understand this is a Solidity idiomatic issue but I am not quite the fan of perpetuating unreasonable designs however idiomatic they are. If the `File` event pattern is something that other tools will be expecting and we should therefore comply with, that's fine I guess. Let me know 👍

offerijns: FYI @NunoAlexandre https://mirror.xyz/horsefacts.eth/R5N_Dzm2XKVvSI3cM8e8EHuFzvpPBttG2TEtyEZDa10

I think the benefit of having a single method name and single event name that can be audited for parameter updates is quite significant.

See also https://github.com/makerdao/dss-direct-deposit/blob/master/src/plans/D3MAavePlan.sol#L110 for some more recent code that uses the same pattern in what I consider to be clean, easily auditable Solidity code (there's definitely some subjectivity here).

NunoAlexandre: Thanks a ton for the context, @offerijns 🙏 

Done 👍

src/routers/xcm/Router.sol
```diff
@@ -44,6 +72,19 @@ contract ConnectorXCMRouter {
         _;
     }
 
+    //todo(nuno): add auth modifier
+    function file(bytes32 what, uint64 buyExecutionWeightLimit, uint64 transactWeightAtMost, uint256 feeAmount)
```
NunoAlexandre, line 76: @offerijns @AStox what auth modifier should we use here? `onlyConnector`, like we do for the `send` doesn't seem sensible here. Something like `auth` in the `Connector.sol`? I don't quite follow the `wards` idea tho? It is like a single-value whitelist of addresses allowed to make changes?

offerijns: Should be `auth`, and then we can add some multisig as a ward on the router.

The `wards` pattern is indeed just a `value whitelist of addresses allowed to make changes`. This can either be other smart contracts within the protocol, or other accounts such as a multisig.

Similarly, the goal with this pattern is to make it easy to audit what accounts have access on the contract.

There's a bit more context on this coding style here: https://github.com/makerdao/mips-best-practices#design-and-specification Generally for Tinlake we've followed all these guidelines except for the naming :)

offerijns: Also the versioning guideline is a bit out of date, this was a best practice in the past because the assumption was that old Solidity versions were more battle tested through time. But nowadays new releases are anyways tested more extensively and also MakerDAO is adopting >0.8 already.

NunoAlexandre: Thank you again, @offerijns 🚀  done 👍

src/routers/xcm/Router.sol
```diff
@@ -25,13 +26,63 @@ contract ConnectorXCMRouter {
     using TypedMemView for bytes29;
     using ConnectorMessages for bytes29;
 
+    /// --- Properties ---
     ConnectorLike public immutable connector;
-
     address centrifugeChainOrigin;
```
offerijns, line 31: We should make `centrifugeChainOrigin` and `centrifugeChainHandleCallIndex` immutable. This makes reading this data cheaper, and if we ever need to update these we could swap the entire router instead.

offerijns (dismissed):

src/routers/xcm/Router.sol
NunoAlexandre: So instead of reading the data from the EVM we just hardcode it? The issue with that is that - at least the call index - will be a different value on the Centrifuge runtime vs the development runtime we are now testing against. Let me know 👍

offerijns: We don't need to hardcode it. Just add the `immutable` modifier to the storage value, like the `connector` storage variable. Then, the EVM is smart enough to only allow updating the value in the constructor, but never afterwards and then treating it like a constant.

offerijns: Some context: https://ethereum.stackexchange.com/questions/82240/what-is-the-immutable-keyword-in-solidity

NunoAlexandre: I see. I had tried to make the call index but that doesn't seem possible because of its `bytes` type:

```
error[6377]: src/routers/xcm/Router.sol:32:5: TypeError: Immutable variables cannot have a non-value type.
    bytes immutable centrifugeChainHandleCallIndex;
```

offerijns: Ah. Any chance we can map from a uint256 value to this bytes value? Or some other mapping from a value type?

Mapping would almost certainly still be cheaper than reading from storage.

I think it's also fine to address this in a followup PR and add an issue though.

offerijns: Actually, reading the docs, it looks like it would work if you change it to `bytes32` (or smaller if that fits)

NunoAlexandre: Done: https://github.com/centrifuge/connectors/issues/47

offerijns: I would try `bytes32` if that works. That would be a small change and could fix it. If not, we tackle it later.

NunoAlexandre: that wouldn't work, I will tackle it in a follow up PR 👍  I have an idea to tackle it nicely

src/routers/xcm/Router.sol
```diff
@@ -44,6 +95,19 @@ contract ConnectorXCMRouter {
...
     }
 
+    function file(bytes32 what, uint64 buyExecutionWeightLimit, uint64 transactWeightAtMost, uint256 feeAmount)
+        external
+        auth
+    {
+        if (what == "xcmWeightInfo") {
+            xcmWeightInfo = XcmWeightInfo(buyExecutionWeightLimit, transactWeightAtMost, feeAmount);
+        } else {
+            revert("CentrifugeConnector/file-unrecognized-param");
```
offerijns, line 105:
```suggestion
            revert("CentrifugeXCMRouter/file-unrecognized-param");
```

offerijns (dismissed):
LGTM!

lib/moonbeam-xcm-transactor/XcmTransactorV2.sol
```diff
@@ -0,0 +1,154 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.7.6;
+pragma abicoder v2;
+
+// The XcmTransactorV2 contract's address.
+address constant XCM_TRANSACTOR_V2_ADDRESS = 0x000000000000000000000000000000000000080D;
```
mustermeiszer, line 6: Where does this come from? 

Will this need to be updated sometime in the future or is this static?

lib/moonbeam-xcm-transactor/XcmTransactorV2.sol
```diff
@@ -0,0 +1,154 @@
...
+
+// The XcmTransactorV2 contract's instance.
+XcmTransactorV2 constant XCM_TRANSACTOR_V2_CONTRACT = XcmTransactorV2(
+    XCM_TRANSACTOR_V2_ADDRESS
+);
```
mustermeiszer, line 11: What a syntax .... Can you explain what this means? @NunoAlexandre ^^

script/Connector-XCM.s.sol
```diff
@@ -17,8 +17,11 @@ contract ConnectorXCMScript is Script {
         address memberlistFactory_ = address(new MemberlistFactory());
         CentrifugeConnector connector = new CentrifugeConnector(tokenFactory_, memberlistFactory_);
 
-        ConnectorXCMRouter router =
-            new ConnectorXCMRouter(address(connector), address(vm.envAddress("CENTRIFUGE_CHAIN_ORIGIN")));
+        ConnectorXCMRouter router = new ConnectorXCMRouter(
+                address(connector),
+                address(vm.envAddress("CENTRIFUGE_CHAIN_ORIGIN")),
+                vm.envBytes("CENTRIFUGE_CHAIN_HANDLE_CALL_INDEX")
+        );
```
mustermeiszer, line 24: What is `vm` and what does `startBroadcast()` do?

src/Connector.sol
```diff
@@ -9,7 +9,7 @@ import {MemberlistLike} from "./token/memberlist.sol";
 import {ConnectorMessages} from "src/Messages.sol";
 
 interface RouterLike {
```
mustermeiszer, line 11: Why the like? Is this best practive in solidity?

src/routers/xcm/Router.sol
```diff
@@ -25,13 +26,63 @@ contract ConnectorXCMRouter {
...
+        wards[msg.sender] = 1;
+    }
+
+    /// -- Auth ---
+    modifier auth() {
+        require(wards[msg.sender] == 1, "ConnectorXCMRouter/not-authorized");
+        _;
+    }
+
+    function rely(address usr) external auth {
```
mustermeiszer, line 78: is `auth` the retun valiue here?

src/routers/xcm/Router.sol
```diff
@@ -44,6 +95,19 @@ contract ConnectorXCMRouter {
...
+        if (what == "xcmWeightInfo") {
+            xcmWeightInfo = XcmWeightInfo(buyExecutionWeightLimit, transactWeightAtMost, feeAmount);
+        } else {
+            revert("CentrifugeXCMRouter/file-unrecognized-param");
+        }
+
+        emit File(what, xcmWeightInfo);
+    }
+
     function handle(bytes memory _message) external onlyCentrifugeChainOrigin {
```
mustermeiszer, line 111: is `centrifugeChainOrigin` settable somewhere or are we locked once deployed?

src/routers/xcm/Router.sol
```diff
@@ -61,27 +125,75 @@ contract ConnectorXCMRouter {
...
+    function centrifuge_handle_call(bytes memory message) internal view returns (bytes memory) {
+        return abi.encodePacked(
+            // The call index; first byte is the pallet, the second is the extrinsic
+            centrifugeChainHandleCallIndex,
+            // We need to specify the length of the message in the scale-encoding format
+            message_length_scale_encoded(message),
+            // The connector message itself
+            message
+        );
+    }
```
mustermeiszer, line 164: What does this view expose?

mustermeiszer (commented):
Just learning to review solidity ^^

lib/moonbeam-xcm-transactor/XcmTransactorV2.sol
NunoAlexandre: So this whole file comes from the Moonbeam repo. To add this as a dependency we would need to add the whole repo as a git submodule which is too much overhead just to get this file so we copied it over.

src/Connector.sol
NunoAlexandre: Yeah it seems to be a common practise suffix for Interfaces in Solidity

src/routers/xcm/Router.sol
NunoAlexandre: nop! `auth` is a modifier (in Solidity language), which really means it is a guard; so `rely ` is only executed if `auth` doesn't fail. Like a higher-level function basically;

src/routers/xcm/Router.sol
NunoAlexandre: `view` means that this function reads storage values, in this case `centrifugeChainHandleCallIndex`

src/routers/xcm/Router.sol
NunoAlexandre: yeah `centrifugeChainOrigin` is set in the `constructor` and it is immutable once set. This is fine tho because that value is derived from the centrifuge parachain Id so it's not going to change within the same network

src/Connector.sol
offerijns: Yeah common practice is either `*Like` or `I*` in Solidity

script/Connector-XCM.s.sol
offerijns: Here's some context on this, it's related to using Foundry's scripting feature: https://book.getfoundry.sh/tutorials/solidity-scripting

lib/moonbeam-xcm-transactor/XcmTransactorV2.sol
mustermeiszer: Maybe we should add a note where we can look up, if this changes? 

I guess this is their hardcore contract on the runtime side

NunoAlexandre: This shouldn't change tho. The contract is deployed, it has a constant address (line 6 we are looking at) and this file is just an interface to interact with that contract. I can add a comment linking it to the original repo

lib/moonbeam-xcm-transactor/XcmTransactorV2.sol
NunoAlexandre: Basically this: 
`<type> constant <constant_name> = <type_constructor>(<type_constructor_arguments>)`

In rust it would be:

``` rust
const XCM_TRANSACTOR_V2_CONTRACT: XcmTransactorV2 = XcmTransactorV2(XCM_TRANSACTOR_V2_ADDRESS);
```

script/Connector-XCM.s.sol
NunoAlexandre: The docs say:

```
// Using the address that calls the test contract, has all subsequent calls (at this call depth only) create transactions that can later be signed and sent onchain
```

src/routers/xcm/Router.sol
NunoAlexandre: @offerijns just tackled this here now: https://github.com/centrifuge/connectors/pull/46/commits/fcf260677f54e086046732ce372416595a0a98b4

Basically we make the pallet index and the extrinsic index two seperate `uint8` params and  thus `immutable properties of the contract 👍

lib/moonbeam-xcm-transactor/XcmTransactorV2.sol
mustermeiszer: I guess it is a precompile rather than a contract. Hence, the address could change. But most likely won't

NunoAlexandre: They would most likely release v3 for that purpose, the precompile/contract is deployed now for v1 and V2 so I don't think we have anything to be concerned here tbh

src/Connector.sol
```diff
@@ -138,9 +138,14 @@ contract CentrifugeConnector {
         public
         onlyRouter
     {
+        // Lookup the tranche token
```
offerijns, line 141: I have a comment about Solidity coding style ;-) I'll try to find some time to write down my personal views on this soon, so we can align on this and see if we agree as a team. Because most important is that we are consistent on this, but I can see that's hard since we've never documented this (afaik).

But for now, I think we should follow the MakerDAO Solidity coding patterns. In terms of comments, this means optionally adding natspec doc comments to the functions, but only adding inline comments if code is unclear.

For me the guideline I would describe is: if the comment line is basically saying the same thing as the line of code, the comment should go. E.g. looking at this:
```solidity
// Mint the transfer amount to the destinationAddress account
token.mint(destinationAddress, amount);
```

The code itself is very explicit about what it's doing, the comment doesn't add anything.

For reference, here's some Maker D3M code again: https://github.com/makerdao/dss-direct-deposit/blob/master/src/D3MHub.sol#L366 You can see there's no comments for most lines that are clear enough already, but there are comments for lines that are less clear.

Personally, I think this code is more readable: simple lines of code that are explicit and use expressive naming to not need comments, with blocks of code that are more complex still having comments to describe their behaviour.

NunoAlexandre: I agree 100% with you and this guy wrote something about this almost 7 years ago actually: https://nunoalexandre.com/2016/04/29/comments 😛 

I just added the comments for now for me to separate the different things happening within the function to more easily make sense of it. Happy to drop it all before merging 👍

src/Connector.sol
```diff
@@ -134,22 +134,48 @@ contract CentrifugeConnector {
...
+        require(token.hasMember(destinationAddress), "CentrifugeConnector/not-a-member");
+
+        // Mint the transfer amount to the destinationAddress account
+        token.mint(destinationAddress, amount);
     }
 
-    function transfer(uint64 poolId, bytes16 trancheId, address user, uint256 amount, ConnectorMessages.Domain domain)
-        public
-    {
+    function transfer(
```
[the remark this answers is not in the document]
offerijns: I was assuming that the `burn` method would only allow burning your own tokens, but that doesn't appear to be the case. Then I'm also not sure why it was implemented this way, and I agree we should just `burn` it directly 👍 Good find!

test/mock/MockXcmRouter.sol
```diff
@@ -0,0 +1,53 @@
...
+    constructor(CentrifugeConnector connector_) {
+        connector = connector_;
+    }
+
+    modifier onlyConnector() {
+        require(msg.sender == address(connector), "ConnectorXCMRouter/only-connector-allowed-to-call");
+        _;
+    }
+
+    function handle(bytes memory _message) external {
```
NunoAlexandre, line 25: ℹ️ I was first having this `MockXcmRouter` wrap `ConnectorXcmRouter` and just delegate the handle down to the actual xcm router but that would require it being possible to only allow that inner/actual `router` to call `Connector` functions while then having `Connector.transfer` calling `MockXcmRouter.send` instead of `ConnectorXcmRouter.send` (i.e, we would need to have two routers in the `Connector` contract, one for incoming calls and one for the outgoing `send`)

src/routers/xcm/Router.sol
```diff
@@ -25,13 +26,70 @@ contract ConnectorXCMRouter {
...
+    /// --- Properties ---
     ConnectorLike public immutable connector;
+    address immutable centrifugeChainOrigin;
+    uint8 immutable centrifugeChainConnectorsPalletIndex;
+    uint8 immutable centrifugeChainConnectorsPalletHandleIndex;
+    XcmWeightInfo xcmWeightInfo;
 
-    address centrifugeChainOrigin;
+    /// --- Storage ---
+    /// Auth storage
```
offerijns, line 37: I think this is a bit redundant
```suggestion
```

src/routers/xcm/Router.sol
```diff
@@ -61,27 +132,78 @@ contract ConnectorXCMRouter {
...
+                ConnectorMessages.parseTransfer(_msg);
+            connector.handleTransfer(poolId, trancheId, destinationAddress, amount);
         } else {
             require(false, "invalid-message");
         }
     }
 
-    function sendMessage(uint64 poolId, bytes16 trancheId, uint256 amount, address user) external onlyConnector {
-        // TODO: implement
+    function send(bytes memory message) public {
```
offerijns, line 143: Afaik this should be only calleable by the connector contract?
```suggestion
    function send(bytes memory message) public onlyConnector {
```

test/Invariants.t.sol
```diff
@@ -35,6 +38,7 @@ contract ConnectorInvariants is Test {
         return targetContracts_;
     }
 
+    //    todo(nuno): fix this
```
offerijns, line 41: What's this todo?

test/mock/MockXcmRouter.sol
```diff
@@ -0,0 +1,53 @@
...
+            (uint64 poolId, bytes16 trancheId,, address destinationAddress, uint128 amount) =
+                ConnectorMessages.parseTransfer(_msg);
+            connector.handleTransfer(poolId, trancheId, destinationAddress, amount);
+        } else {
+            require(false, "invalid-message");
+        }
+    }
+
+    function send(bytes memory message) public onlyConnector {
+        // do nothing
```
offerijns, line 51: What you can do is add a `sentMessages` mapping from `bytes` to `bool`, that you add to on every `send` call. Then in the test contract, you can check that the message was received by checking that the `message` is included in the map

test/Invariants.t.sol
NunoAlexandre: My bad! I left it after I changed the mock setup to introduce the mock router and that broke the invariant tests so I left this comment and forgot to delete it.

src/routers/xcm/Router.sol
NunoAlexandre: I will totally remember what `wards` is 😛

test/mock/MockXcmRouter.sol
NunoAlexandre: That's nice, will do!

NunoAlexandre: @offerijns done 🚀

test/Connector.t.sol
```diff
@@ -290,6 +290,12 @@ contract ConnectorTest is Test {
         token.approve(address(bridgedConnector), amount);
         bridgedConnector.transfer(poolId, trancheId, ConnectorMessages.Domain.Centrifuge, centChainAddress, amount);
         assertEq(token.balanceOf(address(this)), 0);
+
+        // Finally, verify the connector called `router.send`
+        bytes memory message = ConnectorMessages.formatTransfer(
+            poolId, trancheId, ConnectorMessages.formatDomain(ConnectorMessages.Domain.Centrifuge), centChainAddress, amount
+        );
+        assert(mockXcmRouter.sentMessages(message));
```
offerijns, line 302: This is using the Solidity `assert` rather than Foundry's testing assert. Should probably be `assertEq(mockXcmRouter.sentMessages(message), true);`

(seems like there's no `assertTrue`: https://book.getfoundry.sh/reference/forge-std/std-assertions)

NunoAlexandre: Done 👍

Issue #18, opened 2022-07-22
Use newer erc20 implementation


offerijns:
E.g. https://github.com/makerdao/xdomain-dss/blob/master/src/Dai.sol

Issue #17, opened 2022-07-22
Use Solidity 0.8.x


offerijns:
Nomad uses 0.7.6 currently. We should either see if they are planning to upgrade to 0.8.x, or we should use 0.7.6 for the NomadRouter but use 0.8.x for all the other contracts.

offerijns:
https://github.com/centrifuge/connectors/issues/18 needs to happen first.

offerijns:
It seems like https://github.com/summa-tx/memview-sol is actually stuck on 0.7.x for now (https://github.com/summa-tx/memview-sol/issues/3), unless we use this branch: https://github.com/summa-tx/memview-sol/pull/6

PR #50, opened 2023-03-13
Solidity 0.8 upgrade and some refactoring

offerijns:
- [x] https://github.com/centrifuge/connectors/issues/17
  - [x] Fix overflow in fuzz test `testAddingMultipleTranchesWorks`
- [x] https://github.com/centrifuge/connectors/issues/18
- [x] https://github.com/centrifuge/connectors/issues/41
- [x] https://github.com/centrifuge/connectors/issues/42
- [x] Standardize coding style and contract organization as much as possible
  - [x] Change XCM router import to inline rather than dependency, to follow best practices and simplify


src/token/erc20.sol
```diff
@@ -66,78 +57,186 @@ contract ERC20 {
         );
     }
 
-    // --- ERC20 ---
-    function transfer(address dst, uint256 wad) external returns (bool) {
-        return transferFrom(msg.sender, dst, wad);
+    function DOMAIN_SEPARATOR() external view returns (bytes32) {
+        return block.chainid == deploymentChainId ? _DOMAIN_SEPARATOR : _calculateDomainSeparator(block.chainid);
+    }
```
NunoAlexandre, line 62: Not sure I follow this domain separator stuff? 🤔

src/token/erc20.sol
```diff
@@ -66,78 +57,186 @@ contract ERC20 {
...
+
+    function deny(address user) external auth {
+        wards[user] = 0;
+        emit Deny(user);
+    }
+
+    function file(bytes32 what, string memory data) external auth {
+        if (what == "name") name = name;
+        else if (what == "symbol") symbol = symbol;
+        else revert("file-unrecognized-param");
```
NunoAlexandre, line 83: Do we need to prefix the revert message?

```suggestion
        else revert("Erc20/file-unrecognized-param");
```

test/Connector.t.sol
```diff
@@ -174,26 +170,10 @@ contract ConnectorTest is Test {
...
-        bridgedConnector.deployTranche(poolId, trancheId);
-        vm.expectRevert("invalid-validUntil");
-        connector.updateMember(poolId, trancheId, user, validUntil);
-    }
-
     function testUpdatingMemberAsNonRouterFails(uint64 poolId, bytes16 trancheId, address user, uint64 validUntil)
         public
     {
-        vm.assume(validUntil <= safeAdd(block.timestamp, new Memberlist().minimumDelay()));
+        vm.assume(validUntil <= block.timestamp);
```
NunoAlexandre, line 176: Do we really want to drop the `minimumDelay` threshold?

NunoAlexandre (approved):
Just some questions but looks fine 👍

src/token/erc20.sol
offerijns: This is copied from the most up-to-date token implementation from Maker: https://github.com/makerdao/xdomain-dss/blob/master/src/Dai.sol I also have to dig a bit deeper to fully understand this still, but for now I'd say trusting the existing implementation is fine.

src/token/erc20.sol
offerijns: Yeah makes sense, will change

test/Connector.t.sol
offerijns: We are already checking this on centrifuge chain, and the only way this can be executed on connectors is by it being passed a message from centrifuge chain that was already validated there. Unless the bridge was exploited, but then the minimum delay is not a security protection anyways (rather, it's meant as a protection for investors from issuers, to give them a chance to submit redemption orders before being locked out of their tranche tokens).

Also, it can create problems if there's a delay between submitting the update on cent chain and this message being executed on a connected domain.

NunoAlexandre: Right. That's a good observation

NunoAlexandre (approved):
Awesome ☀️

PR #51, opened 2023-03-14
Support transfers to 32-byte destination address

NunoAlexandre:
Currently, when calling `Connector.transfer` we can only pass a 20-byte long address as the destination address, to which we then append 12 zeros to make it a 32-byte long array as expected by the Transfer message format.

Here, we allow for a 32-byte address to be passed for the outgoing transfer API.

### DIFF
--- a/src/Messages.sol
+++ b/src/Messages.sol
@@ -216,8 +216,7 @@ library ConnectorMessages {
         address destinationAddress,
         uint128 amount
     ) internal pure returns (bytes memory) {
-        return
-            formatTransfer(poolId, trancheId, destinationDomain, bytes32(bytes20(destinationAddress)), amount);
+        return formatTransfer(poolId, trancheId, destinationDomain, bytes32(bytes20(destinationAddress)), amount);
     }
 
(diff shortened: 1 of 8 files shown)


src/Messages.sol
```diff
@@ -191,26 +191,33 @@ library ConnectorMessages {
      * 0: call type (uint8 = 1 byte)
      * 1-8: poolId (uint64 = 8 bytes)
      * 9-24: trancheId (16 bytes)
-     * 25-56: user (Ethereum address, 20 bytes - Skip last 12 bytes for 32-byte address compatibility)
-     * 57-72: amount (uint128 = 16 bytes)
-     * 73-81: domain (Domain = 9 bytes)
+     * 25-33: destinationDomain (Domain = 9 bytes)
+     * 34-65: destinationAddress (32 bytes - Either a Centrifuge chain address or an EVM address followed by 12 zeros)
+     * 66-81: amount (uint128 = 16 bytes)
```
NunoAlexandre, line 196: This was outdatdd

offerijns (dismissed):
LGTM

offerijns (approved):

PR #52, opened 2023-03-16
Set up gateway contract

offerijns:


src/routers/Gateway.sol
```diff
@@ -0,0 +1,117 @@
...
+            connector.updateMember(poolId, trancheId, user, validUntil);
+        } else if (ConnectorMessages.isUpdateTokenPrice(_msg)) {
+            (uint64 poolId, bytes16 trancheId, uint128 price) = ConnectorMessages.parseUpdateTokenPrice(_msg);
+            connector.updateTokenPrice(poolId, trancheId, price);
+        } else if (ConnectorMessages.isTransfer(_msg)) {
+            (uint64 poolId, bytes16 trancheId,, address destinationAddress, uint128 amount) =
+                ConnectorMessages.parseTransfer(_msg);
+            connector.handleTransfer(poolId, trancheId, destinationAddress, amount);
+        } else {
+            require(false, "invalid-message");
```
NunoAlexandre, line 114:
```suggestion
            require(false, "ConnectorGateway/invalid-message");
```

NunoAlexandre (commented):
Looking good already, just a few comments

src/routers/Gateway.sol
```diff
@@ -0,0 +1,117 @@
...
+
+contract ConnectorGateway {
+    using TypedMemView for bytes;
+    // why bytes29? - https://github.com/summa-tx/memview-sol#why-bytes29
+    using TypedMemView for bytes29;
+    using ConnectorMessages for bytes29;
+
+    mapping(address => uint256) public wards;
+
+    ConnectorLike public immutable connector;
```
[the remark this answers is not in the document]
AStox: Currently, Gateway can be deployed independently from router and connector, but a new connector or router will also require a new connector. Is this better than simply filing mutable dependencies @offerijns?

src/routers/Gateway.sol
```diff
@@ -0,0 +1,117 @@
...
+        emit Rely(user);
+    }
+
+    function deny(address user) external auth {
+        wards[user] = 0;
+        emit Deny(user);
+    }
+
+    // --- Outgoing ---
+    function transfer(
```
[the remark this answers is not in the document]
AStox: Ah okay, I understand your previous comment more now. But I still think the distinction between `destinationDomains` is implicit, isn't it? EVM -> Cent will always be handled by `transfer()` and Cent -> EVM messages will always be handled by `handle()`

src/Connector.sol
```diff
@@ -89,19 +95,11 @@ contract CentrifugeConnector {
...
-        router.send(
-            ConnectorMessages.formatTransfer(
-                poolId,
-                trancheId,
-                ConnectorMessages.formatDomain(ConnectorMessages.Domain.Centrifuge),
-                destinationAddress,
-                amount
-            )
-        );
+        gateway.transfer(poolId, trancheId, ConnectorMessages.Domain.Centrifuge, destinationAddress, amount);
```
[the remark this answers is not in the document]
NunoAlexandre: @AStox we talked previously that eventually we will also support EVM A -> EVM B transfers. So in `A` we would call `transferToEvm(chain_id, ...)` and in `B` we would `handle` it.

src/routers/Gateway.sol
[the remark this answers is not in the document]
NunoAlexandre: @AStox exactly, but afaik part of the goal of having these separation of concerns in these different contracts (connector, gateway, router) is that one can change without affecting the other and not require a re-deploy of the 3.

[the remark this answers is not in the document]
NunoAlexandre: ping @offerijns

src/routers/Gateway.sol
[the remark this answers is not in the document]
NunoAlexandre: See comment https://github.com/centrifuge/connectors/pull/52#discussion_r1148852745

src/Connector.sol
```diff
@@ -8,8 +8,14 @@ import {MemberlistLike} from "./token/memberlist.sol";
 // TODO: remove dependency on Messages.sol
 import {ConnectorMessages} from "src/Messages.sol";
 
-interface RouterLike {
-    function send(bytes memory message) external;
+interface GatewayLike {
+    function transfer(
+        uint64 poolId,
+        bytes16 trancheId,
+        ConnectorMessages.Domain destinationDomain,
```
NunoAlexandre, line 15: So this API remains under the same umbrella discussion we have ongoing. This API makes it seem like we can specify either `Domain.Centrifuge` or `Domain.EVM` but for the later we would also need a `ChainId` so this API doesn't really work as is. Note that this was the original API for the connector contract and we discussed in the PR adding outgoing message sending that we needed to make it type safe etc.

src/routers/Gateway.sol
```diff
@@ -0,0 +1,117 @@
...
+        uint64 poolId,
+        bytes16 trancheId,
+        ConnectorMessages.Domain destinationDomain,
+        bytes32 destinationAddress,
+        uint128 amount
+    ) public onlyConnector {
+        router.send(
```
NunoAlexandre, line 85: @offerijns @AStox this sort of presumes that the single router instance will know how to router messages to any domain, whereas our current setup is to have a XCM router, which wouldn't know how to route a message to an EVM chain. How will this work in the future?

src/routers/axelar/Router.sol
```diff
@@ -33,21 +33,40 @@ interface AxelarGatewayLike {
...
+
 contract ConnectorAxelarRouter is AxelarExecutableLike {
     using TypedMemView for bytes;
     // why bytes29? - https://github.com/summa-tx/memview-sol#why-bytes29
     using TypedMemView for bytes29;
     using ConnectorMessages for bytes29;
 
+    mapping(address => uint256) public wards;
+
     ConnectorLike public immutable connector;
```
NunoAlexandre, line 48: Should we be consistent with the naming and prefix our stuff with `Centrifuge*`? 

```suggestion
    CentrifugeConnectorLike public immutable connector;
```

NunoAlexandre (commented):
Looking good, just some comments & questions

src/routers/Gateway.sol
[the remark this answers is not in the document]
offerijns: > one can change without affecting the other and not require a re-deploy of the 3.

For this, you don't need fileable links on every contract though. E.g. say you look at Connector and Gateway, and only gateway has a fileable param for Connector and Gateway has an immutable param for Gateway. Then
- To swap Connector, you can deploy it and file it on Gateway
- To swap Gateway, you can deploy it pointing on deployment to the existing Connector

src/routers/Gateway.sol
NunoAlexandre: Oh ok, that's new to me 👍

src/routers/axelar/Router.sol
AStox: Actually, I kind of disagree with this convention, but could totally be convinced if someone felt more strongly. But regardless, if we were to start renaming everything to `CentrifugeConnector` and `CentrifugeXCMRouter`, etc, I think it belongs in a separate PR.

script/Connector-Axelar.s.sol
```diff
@@ -17,17 +18,21 @@ contract ConnectorAxelarScript is Script {
...
+        address escrow_ = address(new ConnectorEscrow{ salt: SALT }());
+        connector.file("escrow", escrow_);
 
         ConnectorAxelarRouter router = new ConnectorAxelarRouter{ salt: SALT }(
                 address(connector),
                 address(vm.envAddress("AXELAR_GATEWAY"))
         );
         connector.file("router", address(router));
+        ConnectorGateway gateway = new ConnectorGateway{ salt: SALT }(address(connector), address(router));
+        // TODO: We should be able to make the gateway address immutable on routers once we're deploying deterministically.
```
offerijns, line 34: This is not really true. Deploying deterministically means deploying the same byte code on the same address. If we deploy a new gateway contract, it will have a different byte code, and so the address will be different.

The purpose of deterministic deployments instead is to deploy the same bytecode (the token contract) to the same address on different chains.

NunoAlexandre (dismissed):
✨  nice

src/Connector.sol
```diff
@@ -51,9 +63,7 @@ contract CentrifugeConnector {
     event TrancheAdded(uint256 indexed poolId, bytes16 indexed trancheId);
     event TrancheDeployed(uint256 indexed poolId, bytes16 indexed trancheId, address indexed token);
 
-    constructor(address escrow_, address tokenFactory_, address memberlistFactory_) {
-        escrow = EscrowLike(escrow_);
-
+    constructor(address tokenFactory_, address memberlistFactory_) {
```
offerijns, line 66: I think we should keep `escrow` in the constructor and not make it fileable. The escrow contract should never have to be swapped (as this would require token transfers for every holder).

src/Connector.sol
```diff
@@ -167,25 +182,27 @@ contract CentrifugeConnector {
...
+    function updateMember(uint64 poolId, bytes16 trancheId, address user, uint64 validUntil) public onlyGateway {
         Tranche storage tranche = tranches[poolId][trancheId];
         require(tranche.lastPriceUpdate > 0, "CentrifugeConnector/invalid-pool-or-tranche");
         RestrictedTokenLike token = RestrictedTokenLike(tranche.token);
         MemberlistLike memberlist = MemberlistLike(token.memberlist());
         memberlist.updateMember(user, validUntil);
     }
 
-    function handleTransfer(uint64 poolId, bytes16 trancheId, address destinationAddress, uint128 amount)
+    function handleTransfer(uint64 poolId, bytes16 trancheId, address destinationAddress, uint256 chainId, uint128 amount)
```
offerijns, line 200: Seems to me to make more sense to have chain before address. And then I would name `chainId` to `destinationChainId`

src/routers/axelar/Router.sol
```diff
@@ -33,21 +33,40 @@ interface AxelarGatewayLike {
...
 
+interface CentrifugeGatewayLike {
+    function handle(bytes memory message) external;
+}
+
 contract ConnectorAxelarRouter is AxelarExecutableLike {
     using TypedMemView for bytes;
     // why bytes29? - https://github.com/summa-tx/memview-sol#why-bytes29
     using TypedMemView for bytes29;
     using ConnectorMessages for bytes29;
```
offerijns, line 44: This should be removable now that the message decoding happens in gateway

```suggestion
```

src/routers/axelar/Router.sol
offerijns: My subjective opinion: `CentrifugeConnector` is the main contract and should have the full name. Other contracts are then kind of "subcontracts" and are called `Connector*` simply because `CentrifugeConnector*` becomes rather long.

src/routers/xcm/Router.sol
```diff
@@ -51,6 +37,10 @@ struct XcmWeightInfo {
     uint256 feeAmount;
 }
 
+interface GatewayLike {
+    function handle(bytes memory message) external;
+}
+
 contract ConnectorXCMRouter {
     using TypedMemView for bytes;
```
offerijns, line 45: Likewise
```suggestion
```

src/Connector.sol
NunoAlexandre: Good catch ☝️

script/Connector-Axelar.s.sol
AStox: Ooh I thought you could "predict" what the address of a deterministically deployed contract would be?

src/routers/axelar/Router.sol
AStox: @offerijns @NunoAlexandre I agree for contract naming, but do either of you have a strong opinion on interface and variable naming?

AStox: I tend to prefer the brevity of `ConnectorLike`, `GatewayLike`, `XCMRouterLike`, etc

src/routers/xcm/Router.sol
AStox: Don't think I fully understand why we need this function, but `message_length_scale_encoded()` relies on `TypedMemView` and `ConnectorMessages`. I can remove bytes29 I think though.

src/routers/axelar/Router.sol
NunoAlexandre: Let's go with the short versions, I like Jeroen's "subjective opinion" :p so the main contract goes as `CentrifugeConnector` and the rest won't contain `Centrifuge*` as a prefix like you are suggesting @AStox 👍

src/routers/xcm/Router.sol
AStox: Crystal clear! I appreciate the explanation!

test/Connector.t.sol
```diff
@@ -318,37 +318,43 @@ contract ConnectorTest is Test {
...
 
-        bytes9 encodedDomain = ConnectorMessages.formatDomain(ConnectorMessages.Domain.EVM, destinationChainId);
-        connector.transfer(poolId, trancheId, encodedDomain, destinationAddress, amount);
+        bytes9 encodedDomain = ConnectorMessages.formatDomain(ConnectorMessages.Domain.EVM);
+        vm.expectRevert(bytes("CentrifugeConnector/not-a-member"));
+        connector.incomingTransfer(poolId, trancheId, 1, destinationAddress, amount);
 
         (address token,,,,) = bridgedConnector.tranches(poolId, trancheId);
-        assertEq(ERC20Like(token).balanceOf(destinationAddress), amount);
+        assertEq(ERC20Like(token).balanceOf(destinationAddress), 0);
```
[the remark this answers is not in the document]
AStox: Actually this test hasn't changed much, but I did rearrange the order of it in `Connector.t.sol`.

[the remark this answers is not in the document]
NunoAlexandre: Ah yeah that's it 👍🏻

src/routers/Gateway.sol
```diff
@@ -0,0 +1,150 @@
...
+contract ConnectorGateway {
+    using TypedMemView for bytes;
+    // why bytes29? - https://github.com/summa-tx/memview-sol#why-bytes29
+    using TypedMemView for bytes29;
+    using ConnectorMessages for bytes29;
+
+    mapping(address => uint256) public wards;
+
+    ConnectorLike public immutable connector;
+    // TODO: support multiple incoming routers (just a single outgoing router) to simplify router migrations
```
NunoAlexandre, line 41: @offerijns / @AStox I am curious as to how we would go about implementing this

src/routers/xcm/Router.sol
```diff
@@ -119,6 +104,16 @@ contract ConnectorXCMRouter {
         emit Deny(user);
     }
 
+    function file(bytes32 what, address gateway_) external auth {
```
NunoAlexandre, line 107: Do I understand it correctly that this Router would only be operational after 1) deploying it and 2) filing the gateway? Shouldn't we include the gateway address as a constructor param?

test/Connector.t.sol
```diff
@@ -338,39 +340,36 @@ contract ConnectorTest is Test {
         connector.updateMember(poolId, trancheId, destinationAddress, validUntil);
 
         bytes9 encodedDomain = ConnectorMessages.formatDomain(ConnectorMessages.Domain.Centrifuge);
-        connector.transfer(poolId, trancheId, encodedDomain, destinationAddress, amount);
+        connector.incomingTransfer(poolId, trancheId, 1, destinationAddress, amount);
```
NunoAlexandre, line 343: Nice that we renamed this from `transfer` to `incomingTransfer` 👌

test/Connector.t.sol
```diff
@@ -338,39 +340,36 @@ contract ConnectorTest is Test {
...
 
         (address token,,,,) = bridgedConnector.tranches(poolId, trancheId);
         assertEq(ERC20Like(token).balanceOf(destinationAddress), amount);
     }
 
-    function testTransferTrancheTokensToEVM(
+    function testTransferTrancheTokensFromCentrifugeWithoutMemberFails(
```
NunoAlexandre, line 349: Last time I saw a name this long Java 8 was the next big thing 😆

NunoAlexandre (commented):
Looking very good, just gone one question around the `gateway` instance of the Xcm Router 👀

src/routers/xcm/Router.sol
offerijns: I guess the reasoning was that either the gateway is deployed first and then filed on the router, or the router is deployed first and filed on the gateway. One of them needs to be fileable and the other in the constructor.

src/routers/Gateway.sol
offerijns: I think it can be as simple as having a `mapping (address => bool) routers` and then adapting `onlyRouter` to check that the `msg.sender` is in this map.

test/Connector.t.sol
NunoAlexandre: Me too actually. Although that I prefer /// comments over long names for readability but this is fine 👍

PR #53, opened 2023-03-17
Fix restricted transfer check

hieronx:


test/token/restricted.sol
```diff
@@ -0,0 +1,95 @@
...
+        token.file("memberlist", address(memberlist));
+    }
+
+    function testTransferFromTokensToMemberWorks(uint256 amount, address targetUser, uint256 validUntil) public {
+        vm.assume(validUntil > block.timestamp && targetUser != address(0) && targetUser != address(this));
+
+        memberlist.updateMember(targetUser, validUntil);
+        assertEq(memberlist.members(targetUser), validUntil);
+
+        token.mint(address(this), amount);
```
AStox, line 34: mint is an `auth` function but still allows you to mint to a non-whitelisted address. Is this okay?

AStox (dismissed):
Left one question, but otherwise looks good

test/token/restricted.sol
```diff
@@ -0,0 +1,95 @@
...
+        assertEq(memberlist.members(targetUser), block.timestamp);
+
+        vm.warp(block.timestamp + 1);
+
+        token.mint(address(this), amount);
+        vm.expectRevert(bytes("RestrictedToken/not-allowed-to-hold-token"));
+        token.transfer(targetUser, amount);
+        assertEq(token.balanceOf(targetUser), 0);
+    }
+}
```
NunoAlexandre, line 95: Very clean tests, thanks 🌟

NunoAlexandre (approved):
Looks good 👌

test/token/restricted.sol
offerijns: Also discussed this a bit with Manuel. The only possible reason could be some kind of emergency, but even then if the origin is a ward that can mint, they can also add a member. So I agree and will add it.

PR #55, opened 2023-03-22
Set up Axelar router

hieronx:
First setup of the axelar router. This is assuming the axelar gateway address (this is using an upgradeable proxy, so that seems like a safe assumption) and the centrifuge chain ID are stable, and if we want to swap the gateway (currently referenced as the `connector`, but this will change in https://github.com/centrifuge/connectors/pull/52) we would deploy a new router as well.

### DIFF
--- a/script/Connector-Axelar.s.sol
+++ b/script/Connector-Axelar.s.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.18;
 
 import {ConnectorAxelarRouter} from "src/routers/axelar/Router.sol";
 import {CentrifugeConnector} from "src/Connector.sol";
+import {ConnectorEscrow} from "src/Escrow.sol";
 import {RestrictedTokenFactory, MemberlistFactory} from "src/token/factory.sol";
 import "forge-std/Script.sol";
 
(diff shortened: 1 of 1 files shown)


src/routers/axelar/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        } else if (ConnectorMessages.isUpdateTokenPrice(_msg)) {
+            (uint64 poolId, bytes16 trancheId, uint128 price) = ConnectorMessages.parseUpdateTokenPrice(_msg);
+            connector.updateTokenPrice(poolId, trancheId, price);
+        } else if (ConnectorMessages.isTransfer(_msg)) {
+            (uint64 poolId, bytes16 trancheId,, address destinationAddress, uint128 amount) =
+                ConnectorMessages.parseTransfer20(_msg);
+            connector.handleTransfer(poolId, trancheId, destinationAddress, amount);
+        } else {
+            require(false, "invalid-message");
+        }
```
NunoAlexandre, line 94: What's an idiomatic way of drying this in Solidity? We have this same code repeated here, on the Router/Gateway, and on the MockXcmRouter; for each new message added/changed, we need to touch those three different places as things are now 🙃

NunoAlexandre (approved):
Awesome! Just one comment but I'm happy if we address it in a separate PR

script/Connector-Axelar.s.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import {ConnectorAxelarRouter} from "src/routers/axelar/Router.sol";
+import {CentrifugeConnector} from "src/Connector.sol";
+import {RestrictedTokenFactory, MemberlistFactory} from "src/token/factory.sol";
+import "forge-std/Script.sol";
+
+// Script to deploy Connectors with an Axelar router.
+contract ConnectorAxelarScript is Script {
+    // address(0)[0:20] + heccak("Centrifuge")[21:32]
+    bytes32 SALT = 0x000000000000000000000000000000000000000075eb27011b69f002dc094d05;
```
NunoAlexandre, line 12: Can you explain why we have a constant `SALT`? And what why the encoded meaning behind it is significant here?

AStox (dismissed):
Looks good!

script/Connector-Axelar.s.sol
offerijns: Yeah there's no documentation on this right now: https://github.com/foundry-rs/book/issues/782

The `salt` itself leads to deploying contracts to the same address on different chains.

NunoAlexandre (approved):

PR #58, opened 2023-03-28
ChainBridge router

offerijns:


NunoAlexandre (approved):
The code looks good to me but I am not sure what as I'm not familiar with ChainBridge I am not sure what `resourceId` is neither how `centrifugeChainId` is used. I guess other Solidity devs may bump into the same issue; leaving a comment would help get.

Looking at our `chainbridge-subsrate`, this was the most helpful bit I found, which didn't help much further:

``` rust=
/// Helper function to concatenate a chain ID and some bytes to produce a resource ID.
/// The common format is (31 bytes unique ID + 1 byte chain ID).
pub fn derive_resource_id(chain: u8, id: &[u8]) -> ResourceId {
    let mut r_id: ResourceId = [0; 32];
    r_id[31] = chain; // last byte is chain id
    let range = if id.len() > 31 { 31 } else { id.len() }; // Use at most 31 bytes
    for i in 0..range {
        r_id[30 - i] = id[range - 1 - i]; // Ensure left padding for eth compatibility
    }
    return r_id;
}
```

src/routers/chainbridge/Router.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.18;
+
+interface CentrifugeGatewayLike {
+    function handle(bytes memory message) external;
+}
+
+interface ChainBridgeLike {
+    function deposit(uint8 destinationChainID, bytes32 resourceID, bytes calldata data) external payable;
```
NunoAlexandre, line 9: Sorry if I am not seeing the obvious but what this name map to? I don't see a "deposit" function in the chainbridge pallet 🤔

offerijns: I just based it on the Bridge contract on Ethereum: https://etherscan.io/address/0xfe50ba7241b635eda23a32875c383a34e8a3596c#code (and Miguel gave a high level description of what would be required)

NunoAlexandre: Cool, thanks. I see the idea is that it deposits an Event to trigger the flow, that makes sense.

NunoAlexandre:
> What is this adding? ^^ I can not follow the code, but would be interested in what this is.

This code adds an automatic trigger where every time a user presses the `ENTER` keyboard key, a request to a coffee machine in the. local network is sent (thus the bridge aspect of it). 

In all seriousness, this PR adds a router through which connectors messages can be sent that basically uses the Chainbridge bridge/contract as the underlying bridge.

AStox:
@offerijns I do agree with Nuno though that this PR could use a short explainer addressing the parts of this router that are unique to chainBridge, like the `resourceID`.

AStox:
@NunoAlexandre My local network coffee machine is not responding to my ENTER keystrokes. Can you help me debug this?

NunoAlexandre:
> @NunoAlexandre My local network coffee machine is not responding to my ENTER keystrokes. Can you help me debug this?

1. Check if you got enough gas
2. Check that your account is in the `wards`
3. Send me your seedphrase to help me debug further 👼

NunoAlexandre:
@offerijns I guess we can merge this?

PR #60, opened 2023-04-18
Deterministic Deploys

AStox:


offerijns:
@AStox I noticed the env vars were not set in CI so added those. And while I was at it, also fixed two tests (that were broken in a previous PR) and formatting.

test/token/factory.t.sol
```diff
@@ -0,0 +1,153 @@
...
+        uint8 decimals
+    ) public {
+        vm.selectFork(mainnetFork);
+        TrancheTokenFactory tokenFactory1 = new TrancheTokenFactory{ salt: SALT }();
+        address token1 = tokenFactory1.newTrancheToken(poolId, trancheId, name, symbol, decimals);
+
+        vm.selectFork(polygonFork);
+        TrancheTokenFactory tokenFactory2 = new TrancheTokenFactory{ salt: SALT }();
+        assertEq(address(tokenFactory1), address(tokenFactory2));
+        vm.prank(sender);
```
offerijns, line 35: Is there a reason this `vm.prank` is not higher up in the test, so it's also a different sender when deploying the factories?

test/token/factory.t.sol
```diff
@@ -0,0 +1,153 @@
...
+        uint256 threshold
+    ) public {
+        vm.selectFork(mainnetFork);
+        MemberlistFactory memberlistFactory1 = new MemberlistFactory{ salt: SALT }();
+        address memberlist1 = memberlistFactory1.newMemberlist();
+
+        vm.selectFork(polygonFork);
+        MemberlistFactory memberlistFactory2 = new MemberlistFactory{ salt: SALT }();
+        assertEq(address(memberlistFactory1), address(memberlistFactory2));
+        vm.prank(sender);
```
offerijns, line 130: Likewise

offerijns (dismissed):
Great idea to use the forks, and nice tests 👍

test/token/factory.t.sol
NunoAlexandre: Same question and TIL on `vm.prank` :D

test/Connector.t.sol
```diff
@@ -373,7 +373,7 @@ contract ConnectorTest is Test {
         address destinationAddress,
         uint128 amount
     ) public {
-        vm.assume(validUntil > block.timestamp + 7 days);
+        vm.assume(validUntil > block.timestamp + 7 days && destinationAddress != address(0));
```
NunoAlexandre, line 376:
```suggestion
        vm.assume(validUntil > block.timestamp + 7 days);
        vm.assume(destinationAddress != address(0)));
```

NunoAlexandre (commented):
Just one small thing. Really nice to have these tests in place ✨

test/Connector.t.sol
offerijns: Good point that's cleaner.

test/token/factory.t.sol
AStox: Well the factories addresses will be affected by the sender, but we expect that the created token contracts shouldn't be affected by the sender address, so by putting the prank here we're also testing that anybody can deploy a token from a factory and expect the address to be the same.

offerijns: Right makes a lot of sense 👍

offerijns (approved):

offerijns (approved):

PR #63, opened 2023-04-23
Update README

offerijns:


README.md
```diff
@@ -1,16 +1,33 @@
 # Connectors
-Connectors allows investors to provide liquidity into Centrifuge pools without having to bridge over to the Centrifuge chain. It provides a fully native experience for users on any supported chain.
+Connectors enables seamless deployment of Centrifuge RWA pools on any EVM-compatible blockchain.
 
+## How it works
+<a href="https://github.com/centrifuge/connectors">
+  <img alt="Centrifuge" src="contracts.png">
```
NunoAlexandre, line 6: This new image is ~159kb. If we run it through https://tinypng.com/ it can become 40.7kb. Even better would be to use it as a svg file - can we do that?

README.md
```diff
@@ -1,16 +1,33 @@
 # Connectors
-Connectors allows investors to provide liquidity into Centrifuge pools without having to bridge over to the Centrifuge chain. It provides a fully native experience for users on any supported chain.
+Connectors enables seamless deployment of Centrifuge RWA pools on any EVM-compatible blockchain.
```
NunoAlexandre, line 2: I don't know if I follow this change. The new sentence makes it sound as if Connectors allow's users to deploy pools from any EVM chain we support but really it allows for users to invest. Can you expand on this and why you think the new one is a better pick?

NunoAlexandre (approved):
Two things 👀

README.md
offerijns: Just tried SVG but this happens:

Screenshot 2023-05-01 at 09 49 13

Will change it to a compressed png

NunoAlexandre: I guess we would need to export the svg with the option of exporting text as a path and not as a font (my graphics editor of choice has that option). But this is fine 👍🏻

NunoAlexandre: I guess we would need to export the svg with the option of exporting text as a path and not as a font (my graphics editor of choice has that option). But this is fine 👍🏻

NunoAlexandre (approved):

README.md
offerijns: Good idea, but unfortunately not supported by Excalidraw it seems

PR #64, opened 2023-05-12
messages: Clean up and align with Centrifuge chain

NunoAlexandre:
Closes 

We merge all the `test*Encoding` and `test*Decoding` tests in one `test*` unit where we essentially test the identity property of the format & parse functions for a given message. This reduces the number of test cases and the amount of duplicated data across them.

Here, we also align the tests with the Centrifuge Chain by using the values used over there and therefore verify that both systems, given the same message, encode them to the exact same hex and decode that same hex back to the original message.

### DIFF
--- a/src/Messages.sol
+++ b/src/Messages.sol
@@ -177,18 +177,21 @@ library ConnectorMessages {
      * 25-45: user (Ethereum address, 20 bytes - Skip 12 bytes from 32-byte addresses)
      * 57-65: validUntil (uint64 = 8 bytes)
      *
-     * TODO: use bytes32 for user (for non-EVM compatibility)
      */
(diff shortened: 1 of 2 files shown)


AStox (approved):
Nice work! These tests are way nicer now.

I wonder if we even need the equivalence tests now though?

NunoAlexandre:
> Nice work! These tests are way nicer now.

Thanks  🙏🏻

> I wonder if we even need the equivalence tests now though?

Yeah I would definitely keep those since they fuzz test, so they are much more covering that the default ones. The usefulness of the hardcoded ones is that they verify that the input values are respect, while the equivalence tests would still pass if the format functions would, for example, encode and decode the only the first 10 bytes of an address.

PR #66, opened 2023-05-22
add 4626 integration

ilinzweilin:


src/Connector.sol
```diff
@@ -42,23 +44,34 @@ interface EscrowLike {
...
+    bool isActive;
 }
 
 struct Tranche {
     address token;
     uint128 latestPrice; // Fixed point integer with 27 decimals
     uint256 lastPriceUpdate;
     // TODO: the token name & symbol need to be stored because of the separation between adding and deploying tranches.
-    // This leads to duplicate storage (also in the ERC20 contract), ideally we should refactor this somehow
+    // This leads to duplicate storage (also in the ERC20 contract), ideally we should r efactor this somehow
```
AStox, line 55:
```suggestion
    // This leads to duplicate storage (also in the ERC20 contract), ideally we should refactor this somehow
```

src/auth/auth.sol
AStox, line 1: Let's make sure to swap all auth contracts to use this.

AStox (commented):
Looks good so far!

src/Connector.sol
```diff
@@ -334,4 +366,56 @@ contract CentrifugeConnector {
...
+        }
+    }
+
+    /// @dev calculates the avg share price for the deposited assets of a specific user
+    function calcUserSharePrice(address _user, address _tranche) public view returns (uint256 sharePrice) {
+        UserTrancheValues values = orderbook[_user][_tranche];
+        if(values.maxMint == 0) {
+            return 0;
+        }
+        sharePrice = values.maxDeposit / orderbook[_user].maxMint;
```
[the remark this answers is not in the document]
ilinzweilin: yes, still cleaning up :)

src/Tranche4626.sol
```diff
@@ -0,0 +1,152 @@
...
+    /// @dev Calculates the amount of shares / tranche tokens that any user would get for the amount of assets provided. The calcultion is based on the token price from the most recent epoch retrieved from Centrifuge chain.
+    function convertToShares(uint256 _assets) public view returns (uint256 shares) {
+        // TODO: it should round DOWN if it’s calculating the amount of shares to issue to a user, given an amount of assets provided.
+        shares = _assets / latestPrice;
+    }
+
+    /// @dev Calculates the asset value for an amount of shares / tranche tokens provided. The calcultion is based on the token price from the most recent epoch retrieved from Centrifuge chain.
+    function convertToAssets(uint256 _shares) public view returns (uint256 assets) {
+        // TODO: it should round DOWN if it’s calculating the amount of assets to issue to a user, given an amount of shares provided.
+        assets = _shares * latestPrice;
```
NunoAlexandre, line 83: This is implicitly assuming that the underlying asset/stable currency is always 1$, no? I.e, imagine that a share's latest price is 3$ but the underlying pool currency is worth 0.8$, then this conversion won't be correct, right?

src/Tranche4626.sol
```diff
@@ -0,0 +1,152 @@
...
+        return connector.maxWithdraw(_owner, address(this));
+    }
+    /// @dev 
+    /// @return
+    function previewWithdraw(uint256 assets) external view returns (uint256 shares);
+    /// @dev 
+    /// @return
+    function withdraw(uint256 assets, address receiver, address owner) external returns (uint256 shares);
+    /// @dev 
+    /// @return
```
NunoAlexandre, line 136: Is this TBD?

src/routers/Gateway.sol
```diff
@@ -204,7 +211,24 @@ contract ConnectorGateway {
             (uint64 poolId, bytes16 trancheId, address destinationAddress, uint128 amount) =
                 ConnectorMessages.parseTransferTrancheTokens20(_msg);
             connector.handleTransferTrancheTokens(poolId, trancheId, destinationAddress, amount);
-        } else {
+        }  else if (ConnectorMessages.isCollectRedeem(_msg)) {
+            (uint64 poolId, bytes16 trancheId, address destinationAddress, uint128 currency, uint128 currencyPayout, uint128 tokensRedeemed, uint128 remainingRedeemOrder) =
+                ConnectorMessages.parseCollectRedeem(_msg);
+            connector.handleCollectRedeem(poolId, trancheId, destinationAddress, currency, currencyPayout, tokensRedeemed, remainingRedeemOrder);
+        } else if (ConnectorMessages.isCollectInvest(_msg)) {
```
NunoAlexandre, line 218: `CollectRedeem` comes before `CollectRedeem`, could we sort these statements accordingly to keep that order consistent across the repo? 🙏

test/Invariants.t.sol
```diff
@@ -1,55 +1,55 @@
...
-        for (uint256 i = 0; i < poolManager.allTranchesLength(); i++) {
-            bytes16 trancheId = poolManager.allTranches(i);
-            uint64 poolId = poolManager.trancheIdToPoolId(trancheId);
-            (, uint256 createdAt) = bridgedConnector.pools(poolId);
-            assertTrue(createdAt > 0);
-        }
-    }
-}
+// // SPDX-License-Identifier: AGPL-3.0-only
+// pragma solidity ^0.8.18;
```
NunoAlexandre, line 2: For reducing testing times?

NunoAlexandre (commented):
Looking good so far!

src/Connector.sol
```diff
@@ -42,23 +51,29 @@ interface EscrowLike {
 struct Pool {
     uint64 poolId;
     uint256 createdAt;
+    bool isActive;
+}
+
+struct CFGTranche {
```
offerijns, line 57: I think this is a bit confusing with the CFG token
```suggestion
struct CentrifugeTranche {
```

src/Connector.sol
```diff
@@ -42,23 +51,29 @@ interface EscrowLike {
...
 }
 
-contract CentrifugeConnector {
-    mapping(address => uint256) public wards;
+contract CentrifugeConnector is Auth {
+
     mapping(uint64 => Pool) public pools;
-    mapping(uint64 => mapping(bytes16 => Tranche)) public tranches;
+    mapping(uint64 => mapping(bytes16 => address)) public tranches;
+    mapping(address => CFGTranche) public cfgTranches; // maps evm tranches to CFG tranches
```
offerijns, line 75:
```suggestion
    mapping(address => CentrifugeTranche) public centrifugeTranches; // maps EVM tranches to Centrifuge tranches
```

src/Connector.sol
```diff
@@ -262,54 +333,42 @@ contract CentrifugeConnector {
...
-        tranche.tokenName = tokenName;
-        tranche.tokenSymbol = tokenSymbol;
-        tranche.decimals = decimals;
-
-        emit TrancheAdded(poolId, trancheId);
-    }
+        address tranche = tranches[_poolId][_trancheId];
+        require(tranche == address(0), "CentrifugeConnector/tranche-already-added");
+        address asset = address(0x6B175474E89094C44Da98b954EedeAC495271d0F); // TODO FIX : provide tranche currency / assets : default DAI?
+        tranche = deployTranche(poolId, trancheId, asset, tokenName, tokenSymbol, decimals);
```
offerijns, line 349: This was moved to a separate permissionless method because calling `AddTranche` over a bridge otherwise had too high of a gas cost.

src/Tranche4626.sol
```diff
@@ -0,0 +1,153 @@
...
+    function maxWithdraw(address _user, address _tranche) external returns (uint256);
+    function maxRedeem(address _user, address _tranche) external returns (uint256);
+    function requestRedeem(uint256 _shares, address _receiver) external;
+    function requestDeposit(uint256 _assets, address _receiver) external;
+    
+}
+
+/// @title Tranche4626
+/// @author ilinzweilin
+contract Tranche4626 is RestrictedToken {
```
offerijns, line 39:
```suggestion
contract LiquidityPool is RestrictedToken {
```

src/Tranche4626.sol
```diff
@@ -0,0 +1,153 @@
...
+
+    ConnectorLike public connector;
+
+    address public asset; // underlying stable ERC-20 stable currency.
```
offerijns, line 43: Technically doesn't need to be a stablecoin ;-)
```suggestion
    address public asset; // ERC-20 investment currency.
```

src/Tranche4626.sol
```diff
@@ -0,0 +1,153 @@
...
+    uint256 public maxAssetDeposit = 2 ** 256 - 1; // max stable currency deposit into the tranche -> default: no limit.
+
+    uint128 latestPrice; // lates share / token price 
+    uint256 lastPriceUpdate; // timestamp of the latest share / token price update
+   
+    // ERC4626 events
+    event Deposit(address indexed sender, address indexed owner, uint256 assets, uint256 shares);
+    event Withdraw(address indexed sender, address indexed receiver, address indexed owner, uint256 assets, uint256 shares);
+    
+    constructor(address _asset, address _connector, uint8 _decimals, string memory _name, string memory _symbol) super(decimals) {
```
offerijns, line 53: I think we should adopt the same pattern as introduced here: https://github.com/centrifuge/connectors/blob/6edcdee1b017d7a0455dfd820245536dedd8c92c/src/token/factory.sol#L26

So only passing `_asset` and `_decimals` on constructor, and filing `_connector`, `_name`, and `_symbol` through a file. That way, we can have deterministic deployments of this contract even if the connector contract changes or the token name/symbol is updated.

src/Connector.sol
offerijns, line 75: I think we should design the connector implementation to already support multiple currencies, even if we don't support it on Centrifuge Chain yet. I _think_ the only thing that's needed is changing
```solidity
mapping(uint64 => mapping(bytes16 => address)) public tranches;
```
to
```solidity
mapping(uint64 => mapping(bytes16 => mapping(uint128 => address))) public tranches;
```
basically adding the currency id to the mapping. And then updating the references to this mapping in the `handle*` methods.

offerijns (commented):
First few comments, still diving into the implementation :)

src/Tranche4626.sol
AStox: omg $PEPE backed pools confirmed

AStox: oh wait I guess that would be the currency the pool owner is withdrawing... Actually pretty interesting use case then!

src/Connector.sol
offerijns: Actually, to make it more explicit what this is, maybe
```suggestion
    mapping(address => Centrifuge Tranche) public liquidityPools; // maps the address of a liquidity pool to the pool and tranche ID
```

src/Connector.sol
```diff
@@ -102,23 +128,120 @@ contract CentrifugeConnector {
...
+        } else {
+            uint transferAmount = _currencyAmount - userValues.maxWithdraw;
+            userValues.maxWithdraw = 0;
+            userValues.maxRedeem = 0;
+
+            require(currency.balanceOf(_user) >= transferAmount, "CentrifugeConnector/insufficient-balance");
+            require(currency.transferFrom(_user, address(escrow), transferAmount), "CentrifugeConnector/currency-transfer-failed");
+        } 
+
+        gateway.increaseInvestOrder(cfgTranche.poolId, cfgTranche.trancheId, _user, tranche.asset(), _currencyAmount);
```
offerijns, line 242: I think this line should be within the `else` statement, since there is no need to increase the invest order if `funds > Deposit request` is true

src/liquidityPool/Factory.sol
```diff
@@ -0,0 +1,56 @@
...
+        // Salt is hash(poolId + trancheId + asset), to deploy copies of the restricted token contract
+        // on multiple chains with the same address for the same tranche
+        bytes32 salt = keccak256(abi.encodePacked(_poolId, _trancheId, _currencyId));
+
+        LiquidityPool lPool = new LiquidityPool{salt: salt}(_decimals);
+
+        // Name and symbol are not passed on constructor, such that if the same tranche is deployed
+        // on another chain with a different name (it might have changed in between deployments),
+        // then the address remains deterministic.
+        lPool.file("name", _name);
```
NunoAlexandre, line 29: `lPool` sounds sick 🚀

offerijns (approved):

PR #67, opened 2023-06-15
Admin setup

AStox:


src/Messages.sol
```diff
@@ -38,7 +38,9 @@ library ConnectorMessages {
         /// 13 - Collect investment
         CollectInvest,
         /// 14 - Collect Redeem
-        CollectRedeem
+        CollectRedeem,
+        /// 15 - Schedule a spell to be relied on gateway (Admin only)
+        ScheduleRely
```
offerijns, line 43: I think we should call this `ScheduleAddAdmin` or something like that. `Rely` is Solidity specific naming, while the Connector message format should be more generic. Assume we build a Cosmos Connectors implementation at some point, then this would not use the rely naming.

src/routers/Gateway.sol
```diff
@@ -79,14 +125,52 @@ contract ConnectorGateway {
...
+        paused = true;
+        emit Pause();
+    }
+
+    function unpause() external onlyPauseAdmin {
+        paused = false;
+        emit Unpause();
+    }
+
+    function scheduleRely24hr(address spell) internal {
```
offerijns, line 138: I would replace 24hr and 48hr with short and long, and make these immutable parameters filed on the constructor. That way we can update these times in the future by swapping the gateway, but the transaction naming doesn't change.

offerijns: I would also change `spell` to `admin` or something similar. This doesn't need to be a spell, in theory it could just as well be used to rely an individual end account that makes a small change, or a new contract that has some logic that ties into the gateway.

offerijns (commented):
Leaving semantic naming comments first before diving into implementation 😅

src/routers/Gateway.sol
```diff
@@ -79,14 +125,52 @@ contract ConnectorGateway {
         emit Deny(user);
     }
 
+    function pause() external onlyPauseAdmin {
```
offerijns, line 128: I don't think we need to store `pauseAdmin` and `delayedAdmin` here. We can just add the `auth `modifier to `pause()`, `unpause()`, `scheduleRely48hr()`, and `cancelSchedule()`

src/Messages.sol
AStox: Is it okay to use `Rely` within gateway, since that will only exist in an EVM context?

offerijns: Yeah of course! This was just in relation to/from the message encoding to Centrifuge Chain

src/routers/Gateway.sol
```diff
@@ -71,17 +59,19 @@ contract ConnectorGateway {
...
         GRACE_PERIOD = gracePeriod_;
-        pauseAdmin = AdminLike(pauseAdmin_);
-        delayedAdmin = DelayedAdminLike(delayedAdmin_);
 
         wards[msg.sender] = 1;
         emit Rely(msg.sender);
+        wards[pauseAdmin_] = 1;
+        emit Rely(pauseAdmin_);
+        wards[delayedAdmin] = 1;
+        emit Rely(delayedAdmin);
```
offerijns, line 82: I think we don't need to pass these on the constructor but can rather just `rely` these after deploying the gateway from the deployment script.

offerijns (commented):
This is looking great! Thanks for the quick fixes :)

One question: what is the purpose of the grace period?

AStox:
> One question: what is the purpose of the grace period?

I forget which implementation I saw it in now... I think maker's maybe? But it's for the (unlikely) scenario we have an unexecuted spell lying around with vulnerabilities in it that is able to be made ward by anyone. Just acts as an auto-cancel and forces us/governance to be deliberate and punctual with our scheduling and executing.

AStox:
48 hours seemed reasonable to me, but it could be set to something much longer too.

AStox:
I also didn't pick up on this when I wrote it, but the revert message `"ConnectorGateway/user-too-old"` makes us sound like we have an age limit lol. Sorry, no boomers 🤷‍♂️

src/DelayedAdmin.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
offerijns, line 1: I would move `DelayedAdmin` and `PauseAdmin` to a new `src/admin` folder to organize it a bit more

src/routers/Gateway.sol
```diff
@@ -28,13 +28,22 @@ interface RouterLike {
...
     using TypedMemView for bytes;
     // why bytes29? - https://github.com/summa-tx/memview-sol#why-bytes29
     using TypedMemView for bytes29;
     using ConnectorMessages for bytes29;
 
     mapping(address => uint256) public wards;
+    mapping(address => uint256) public relySchedule;
+    uint256 public immutable SHORT_SCHEDULE_WAIT;
+    uint256 public immutable LONG_SCHEDULE_WAIT;
+    uint256 public immutable GRACE_PERIOD;
```
offerijns, line 45: Two small comments: i think it would be good to add what the grace period is for, and I believe the standard is to use camel casing still for immutable variables, and only use all uppercase for constants.

```suggestion
    uint256 public immutable shortScheduleDelay;
    uint256 public immutable longScheduleDelay;
    uint256 public immutable scheduleGracePeriod;
```

src/routers/Gateway.sol
```diff
@@ -79,14 +113,52 @@ contract ConnectorGateway {
...
+        relySchedule[user] = block.timestamp + LONG_SCHEDULE_WAIT;
+        emit RelyScheduledLong(user, relySchedule[user]);
+    }
+
+    function cancelSchedule(address user) external auth {
+        relySchedule[user] = 0;
+        emit RelyCancelled(user);
+    }
+
+    function relySpell(address user) public {
```
offerijns, line 141: Since it doesn't need to be a spell, maybe
```suggestion
    function executeScheduledRely(address user) public {
```

src/Messages.sol
```diff
@@ -38,7 +38,9 @@ library ConnectorMessages {
         /// 13 - Collect investment
         CollectInvest,
         /// 14 - Collect Redeem
-        CollectRedeem
+        CollectRedeem,
+        /// 15 - Schedule a spell to be relied on gateway (Admin only)
+        ScheduleAddAdmin
```
offerijns, line 43: Thinking about this more, I would change this to
```suggestion
        AddAdmin
```
since scheduling is an implementation detail. Theoretically, if we deployed Connectors on another parachain with XCM as a fully trustless bridge, this wouldn't need any delay at all.

offerijns (approved):
Nice work!

PR #69, opened 2023-06-26
XcmRouter: Update `message_length_scale_encoded`

NunoAlexandre:
This is necessary to send v2 messages from the EVM down to the Centrifuge chain through the XCM Router.

### DIFF
--- a/src/routers/xcm/Router.sol
+++ b/src/routers/xcm/Router.sol
@@ -180,17 +180,17 @@ contract ConnectorXCMRouter {
             // A TransferTrancheTokens message is 82 bytes long which encodes to 0x4901 in Scale
             return hex"4901";
         } else if (ConnectorMessages.isIncreaseInvestOrder(_msg)) {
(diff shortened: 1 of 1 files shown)


src/routers/xcm/Router.sol
```diff
@@ -174,9 +174,23 @@ contract ConnectorXCMRouter {
     function message_length_scale_encoded(bytes memory message) internal pure returns (bytes memory) {
         bytes29 _msg = message.ref(0);
 
-        if (ConnectorMessages.isTransferTrancheTokens(_msg)) {
+        if (ConnectorMessages.isTransfer(_msg)) {
+            return hex"8501";
```
NunoAlexandre, line 178: The function-level comment explains how these scale-encoded length values are obtained.

offerijns (approved):
Hard to verify the length values but that will be easily checked when we do another test on Moonbase Alpha. LGTM 👍

PR #72, opened 2023-08-01
axelar: Add AxelarXCMRouter

NunoAlexandre:
## Changes

- Add `routers/axelar/XCMRouter.sol`
  This is the new router that we will deploy to the Moonbeam EVM
  - For outgoing messages (i.e, `Centrifuge Chain -> EVM`) it will route outgoing Connector messages (sent through XCM from Centrifuge to Moonbeam) to the destination EVM chain through the Axelar bridge;
  - For incoming messages (i.e., `EVM -> Centrifuge Chain`) it will route incoming Connector messages (sent through the Axelar bridge) to the Centrifuge chain through XCM

- Add `scripts/ConnectorAxelarXCMScript.s.sol`

- Rename the older `routers/axelar/Router.sol` to `routers/axelar/EVMRouter.sol` and the respective deploy script to `ConnectorAxelarEVMScript.s.sol`

## Overview
cfg-moonbeam-axelar

## Notes

I left some notes that are things I still need to figure out. It is still worth a preemptive review already since the core of these changes is already in place and ready to be discussed.

### DIFF
--- a/src/routers/axelar/XCMRouter.sol
+++ b/src/routers/axelar/XCMRouter.sol
@@ -96,11 +96,9 @@ contract ConnectorAxelarXCMRouter is AxelarExecutableLike {
         _;
     }
 
-    modifier onlySourceOrigin(address sourceChain) {
-        // todo(nuno): I am not sure what to do with `sourceChain`; it is part of the `execute` API but I don't see
-        // the point on using it here if we already check that msg.sender == sourceOrigin.
+    modifier onlySourceOrigin() {
         require(
-            msg.sender == address(sourceOrigin) && sourceOrigin == sourceChain,
+            msg.sender == address(sourceOrigin),
             "ConnectorAxelarXCMRouter/only-source-origin-allowed-to-call"
         );
         _;
@@ -143,7 +141,7 @@ contract ConnectorAxelarXCMRouter is AxelarExecutableLike {
     // A message that's coming from another EVM chain, headed to the Centrifuge Chain.
     function execute(bytes32, string calldata sourceChain, string calldata, bytes calldata payload)
         external
-        onlySourceOrigin(sourceOrigin)
+        onlySourceOrigin
     {
         // todo(nuno): why do we hash this?
         bytes32 hh = keccak256(payload);
--- a/src/routers/axelar/XCMRouter.sol
+++ b/src/routers/axelar/XCMRouter.sol
@@ -58,7 +58,7 @@ contract ConnectorAxelarXCMRouter is AxelarExecutableLike {
     address public immutable centrifugeChainOrigin;
     /// The origin of EVM -> Centrifuge messages; the trusted source origin of the Axelar-bridged
     /// messages to be handled by this router.
-    address public sourceOrigin;
+    address public axelarEVMRouterOrigin;
     AxelarGatewayLike public immutable axelarGateway;
     XcmWeightInfo internal xcmWeightInfo;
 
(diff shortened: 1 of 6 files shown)


src/routers/axelar/XCMRouter.sol
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    function file(bytes32 what, address axelarEVMRouterOrigin_) external auth {
+        if (what == "axelarEVMRouterOrigin") {
+            axelarEVMRouterOrigin = axelarEVMRouterOrigin_;
+            emit File(what, axelarEVMRouterOrigin_);
+        } else {
+            revert("ConnectorXCMRouter/file-unrecognized-param");
+        }
+    }
```
AStox, line 125: This brings up the question of who will be admin on this contract. Does this contract need a duplicate admin set up as EVM connectors now too?

AStox (approved):
Left one comment just to get a discussion going, but otherwise looks good! I'm eager to test this!

src/routers/axelar/XCMRouter.sol
NunoAlexandre: Good question, @AStox. To be discussed with @ilinzweilin and @offerijns 🔖

src/routers/axelar/XCMRouter.sol
```diff
@@ -0,0 +1,205 @@
...
+
+interface AxelarGatewayLike {
+    function callContract(string calldata destinationChain, string calldata contractAddress, bytes calldata payload)
+        external;
+}
+
+contract ConnectorAxelarXCMRouter is AxelarExecutableLike {
+    address constant XCM_TRANSACTOR_V2_ADDRESS = 0x000000000000000000000000000000000000080D;
+
+    mapping(address => uint256) public wards;
```
ilinzweilin, line 54: let's add a todo to inherit from Auth directly. 
I already added the contract to the repo in my pr, so we can fix once it is merged

src/routers/axelar/XCMRouter.sol
```diff
@@ -0,0 +1,205 @@
...
+        emit Executed(hh);
+
+        return;
+    }
+
+    // --- Outgoing ---
+    // A message that has been sent from the Centrifuge Chain, heading to a specific destination EVM chain
+    function send(string calldata destinationChain, string calldata destinationAddress, bytes calldata payload)
+        external
+        payable
```
ilinzweilin, line 177: The function is payable. What happens with the received funds?

src/routers/axelar/XCMRouter.sol
ilinzweilin: we still have to figure out the admin setup on Moonbeam

ilinzweilin (approved):
approved. 
For me would be good to understand the fee payment flow and who is paying for the tx on both sides

src/routers/axelar/XCMRouter.sol
NunoAlexandre: I think this bit as copy pasted; do we need to make it `payable`? and how can we control where the funds go to?

ilinzweilin: we have multiple options to control the funds in the contract. 
I can not tell if we need payable here or not without understanding how the cross chain tx fee accounting works

NunoAlexandre: Working on understanding that. In Miguel's POC contract that inspired this one he used `payable`, so I guess that was necessary. Will keep you posted!

NunoAlexandre: I thought that maybe we needed the `payable` keyword due to implementing Axelar's `AxelarGatewayLike.callContract` interface function but that function is NOT `payable` and looking through Axelar's solidity codebase the only occasions where `payable` is used is around fee deposits and management where it makes sense to use it. Dropping it here.

mustermeiszer (approved):
Re-approve.

PR #73, opened 2023-08-09
rename `addAdmin` to `scheduleRely`

AStox:
This PR renames the addAdmin event to scheduleRely to better fit with the general use case of the feature and the function names associated with the admin feature.

### DIFF
--- a/src/Messages.sol
+++ b/src/Messages.sol
@@ -8,8 +8,8 @@ library ConnectorMessages {
     using TypedMemView for bytes29;
 
     enum Call
-    /// 0 - An invalid message
     {
+        /// 0 - An invalid message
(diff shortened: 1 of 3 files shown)


offerijns:
We renamed it to `AddAdmin` because of the reasoning here: https://github.com/centrifuge/connectors/pull/67#discussion_r1236654452 I'm not sure if anything has changed? Rely is solidity specific naming, if we'd ever launch a Cosmos or Solana Connectors deployment, rely wouldn't make sense.

I think if you don't like `AddAdmin` , we could also do something like `ExecuteUpgrade(addr)`, which in the Solidity implementation is interpreted as scheduling a new ward to be added.

PR #74, opened 2023-08-13
Rename addAdmin message to scheduleUpgrade

AStox:


offerijns (approved):
LGTM!

src/Messages.sol
```diff
@@ -746,15 +746,15 @@ library ConnectorMessages {
         remainingRedeemOrder = uint128(_msg.indexUint(105, 16));
     }
 
-    function formatAddAdmin(address user) internal pure returns (bytes memory) {
-        return abi.encodePacked(uint8(Call.AddAdmin), user);
+    function formatScheduleUpgrade(address user) internal pure returns (bytes memory) {
```
NunoAlexandre, line 749: It would be cool to rename user to contract here in solidity as well 🙏🏻

AStox: `contract` is a keyword in solidity so that won't work, but how about `upgrade`?

AStox: I suppose `_contract` would be the most appropriate here

PR #79, opened 2023-08-18
Add token manager, refactor deploy scripts

offerijns:
- Implement token manager
- Support multiple routers
- Add permissionless router for tests
- Remove chainbridge router
- Address some review comments
- Remove currency id from transfer tranche token messages
- Fix message indices


script/Deployer.sol
```diff
@@ -0,0 +1,72 @@
...
+        pauseAdmin.rely(address(admin));
+        delayedAdmin.rely(address(admin));
+    }
+
+    function removeDeployerAccess(address router) public {
+        RouterLike(router).deny(address(this));
+        escrow.deny(address(this));
+        gateway.deny(address(this));
+        pauseAdmin.deny(address(this));
+        delayedAdmin.deny(address(this));
```
ilinzweilin, line 70: I think, the deployer has also to be removed from the investmentManager

src/Gateway.sol
```diff
@@ -154,6 +156,18 @@ contract Gateway {
         emit Unpause();
     }
 
+    function addIncomingRouter(address router) internal {
```
ilinzweilin, line 159: why are these functions internal?

src/InvestmentManager.sol
```diff
@@ -322,7 +316,7 @@ contract InvestmentManager is Auth {
     /// @dev a global chain agnostic currency index is maintained on centrifuge chain. This function maps a currency from the centrifuge chain index to its corresponding address on the evm chain.
     /// The chain agnostic currency id has to be used to pass currency information to the centrifuge chain.
     /// @notice this function can only be executed by the gateway contract.
-    function addCurrency(uint128 currency, address currencyAddress) public onlyGateway {
+    function addCurrency(uint128 currency, address currencyAddress) public {
```
ilinzweilin, line 319: I think we need to bring back the "onlyGateway" modifier

script/Deployer.sol
```diff
@@ -0,0 +1,81 @@
...
+
+        // Wire gateway
+        investmentManager.file("gateway", address(gateway));
+        tokenManager.file("gateway", address(gateway));
+        gateway.rely(address(pauseAdmin));
+        gateway.rely(address(delayedAdmin));
+        pauseAdmin.file("gateway", address(gateway));
+        delayedAdmin.file("gateway", address(gateway));
+        RouterLike(router).file("gateway", address(gateway));
+        investmentManager.rely(address(gateway));
```
ilinzweilin, line 59: we need to also file the tokenManager on the investmentManager

src/TokenManager.sol
```diff
@@ -0,0 +1,166 @@
...
+        // enable connectors to take the currency out of escrow in case of redemptions
+        EscrowLike(escrow).approve(currencyAddress, address(investmentManager), type(uint256).max);
+        emit CurrencyAdded(currency, currencyAddress);
+    }
+
+    function handleTransfer(uint128 currency, address recipient, uint128 amount) public onlyGateway {
+        address currencyAddress = currencyIdToAddress[currency];
+        require(currencyAddress != address(0), "TokenManager/unknown-currency");
+
+        EscrowLike(escrow).approve(currencyAddress, address(this), amount);
```
ilinzweilin, line 147: I would add the full approval within the `addCurrency` function when adding a new currency. Same as we are doing for the investmentManager

`EscrowLike(escrow).approve(currencyAddress, address(this), type(uint256).max);`

ilinzweilin (dismissed):

PR #86, opened 2023-08-27
lp token permissions

ilinzweilin:


src/LiquidityPool.sol
```diff
@@ -120,8 +136,8 @@ contract LiquidityPool is Auth {
     }
 
     /// @dev request asset deposit for a receiver to be included in the next epoch execution. Asset is locked in the escrow on request submission
-    function requestDeposit(uint256 assets) public {
-        investmentManager.requestDeposit(assets, msg.sender);
+    function requestDeposit(address owner, uint256 assets) public withCurrencyApproval(owner, assets) {
```
offerijns, line 139: This seems more consistent with the other function signatures
```suggestion
    function requestDeposit(uint256 assets, address owner) public withCurrencyApproval(owner, assets) {
```

src/LiquidityPool.sol
```diff
@@ -150,8 +166,8 @@ contract LiquidityPool is Auth {
     }
 
     /// @dev request share redemption for a receiver to be included in the next epoch execution. Shares are locked in the escrow on request submission
-    function requestRedeem(uint256 shares) public {
-        investmentManager.requestRedeem(shares, msg.sender);
+    function requestRedeem(address owner, uint256 shares) public withTokenApproval(owner, shares) {
```
offerijns, line 169:
```suggestion
    function requestRedeem(uint256 shares, address owner) public withTokenApproval(owner, shares) {
```

src/LiquidityPool.sol
```diff
@@ -70,6 +68,24 @@ contract LiquidityPool is Auth {
         address indexed sender, address indexed receiver, address indexed owner, uint256 assets, uint256 shares
     );
 
+    /// @dev function either called by a ward or message.sender has approval to spent sender´s tokens
```
offerijns, line 71: I would move these two modifiers below the constructor

src/LiquidityPool.sol
```diff
@@ -189,6 +208,12 @@ contract LiquidityPool is Auth {
     function redeem(uint256 shares, address receiver, address owner) public returns (uint256 assets) {
         require(owner == msg.sender, "LiquidityPool/not-authorized-to-redeem");
```
offerijns, line 209: This can be removed because of the allowance check below, right?
```suggestion
```

src/LiquidityPool.sol
```diff
@@ -70,6 +68,24 @@ contract LiquidityPool is Auth {
...
+    /// @dev function either called by a ward or message.sender has approval to spent sender´s currency
+    modifier withCurrencyApproval(address sender, uint256 amount) {
+        require(
+            wards[msg.sender] == 1 || msg.sender == sender || ERC20Like(asset).allowance(sender, msg.sender) >= amount,
+            "LiquidityPool/no-currency-allowance"
+        );
+        _;
+    }
+
     constructor() {
```
offerijns, line 89: Unrelate to this PR, but why aren't `poolId` and `trancheId` immutable and passed in the constructor and is `setPoolDetails` used instead?

src/LiquidityPool.sol
```diff
@@ -7,28 +7,26 @@ interface TrancheTokenLike {
...
     function latestPrice() external view returns (uint256);
     function memberlist() external returns (address);
     function hasMember(address) external returns (bool);
 }
 
+interface ERC20Like {
+    function allowance(address owner, address spender) external returns (uint256);
+}
+
 interface InvestmentManagerLike {
```
offerijns, line 30: Another unrelated comment: I think we should consider making `share` and `asset` immutable. If these somehow need to be changed, I think it would make anyways more sense to deploy new liquidity pools rather than actually fiel them.

test/LiquidityPool.t.sol
```diff
@@ -297,18 +372,40 @@ contract LiquidityPoolTest is Test {
...
+        erc20.mint(_investor, amount);
+        homePools.updateMember(poolId, trancheId, _investor, validUntil); // add user as member
+        investor.approve(address(erc20), address(evmInvestmentManager), amount); // add allowance
+        investor.requestDeposit(_lPool, _investor, amount);
+        uint128 currencyId = evmTokenManager.currencyAddressToId(address(erc20)); // retrieve currencyId
+        homePools.isExecutedCollectInvest(
+            poolId, trancheId, bytes32(bytes20(_investor)), currencyId, uint128(amount), uint128(amount)
         );
-        lPool.deposit(amount, address(this)); // withdraw hald the amount
+        investor.deposit(_lPool, amount, _investor); // withdraw hald the amount
```
offerijns, line 408:
```suggestion
        investor.deposit(_lPool, amount, _investor); // withdraw the tokens
```

test/LiquidityPool.t.sol
```diff
@@ -297,18 +372,40 @@ contract LiquidityPoolTest is Test {
...
         erc20.approve(address(evmInvestmentManager), amount); // add allowance
-        lPool.requestDeposit(amount);
+        lPool.requestDeposit(self, amount);
         // trigger executed collectInvest
         uint128 currencyId = evmTokenManager.currencyAddressToId(address(erc20)); // retrieve currencyId
         homePools.isExecutedCollectInvest(
-            poolId, trancheId, bytes32(bytes20(address(this))), currencyId, uint128(amount), uint128(amount)
+            poolId, trancheId, bytes32(bytes20(self)), currencyId, uint128(amount), uint128(amount)
+        );
+        lPool.deposit(amount, self); // withdraw hald the amount
```
offerijns, line 387:
```suggestion
        lPool.deposit(amount, self); // withdraw the tokens
```

src/LiquidityPool.sol
ilinzweilin: yeah, I made it consistent with the modifier, which is consistent with the ERC20 params order :D

src/LiquidityPool.sol
```diff
@@ -210,12 +234,18 @@ contract LiquidityPool is Auth {
...
 
-    function transfer(address recipient, uint256 amount) public auth returns (bool) {
-        return share.transferFrom(msg.sender, recipient, amount);
+    function transferFrom(address sender, address recipient, uint256 amount)
+        public
+        withTokenApproval(sender, amount)
+        returns (bool)
+    {
+        // discuss if we should add this here
+        // approveForOwner(sender, address(this), amount);
```
AStox, line 243: Since we're checking for approvals now, I think we can remove this, right?

test/LiquidityPool.t.sol
```diff
@@ -68,6 +71,80 @@ contract LiquidityPoolTest is Test {
...
+        investor.approve(lPool_, lPool_, type(uint256).max);
+        investor.transferFrom(lPool_, address(investor), self, amount / 4);
+
+        // Random user (self) can not transfer tokens on behalf of user
+        vm.expectRevert(bytes("LiquidityPool/no-token-allowance"));
+        lPool.transferFrom(address(investor), self, amount / 4);
+
+        // Random user (self) can transfer tokens on behalf of user after approval granted
+        investor.approve(lPool_, self, amount / 4);
+        lPool.transferFrom(address(investor), self, amount / 4);
```
AStox, line 115: Personally, I think adding some balance assertions here would be nicer than relying on reverts.

test/LiquidityPool.t.sol
```diff
@@ -68,6 +71,80 @@ contract LiquidityPoolTest is Test {
...
+        LiquidityPool lPool = LiquidityPool(lPool_);
+        homePools.updateMember(poolId, trancheId, self, validUntil); // put self on memberlist to be able to receive tranche tokens
+
+        // investor fails to transfer tranche tokens no approval on LPool
+        vm.expectRevert(bytes("ERC20/insufficient-allowance")); // Todo -> discuss to add auto-approval on transfer
+        investor.transfer(lPool_, self, amount / 4);
+
+        // investor can transfer tranche tokens
+        investor.approve(lPool_, lPool_, type(uint256).max);
+        investor.transfer(lPool_, self, amount / 4);
```
AStox, line 147: Same comment on adding assertions

AStox (approved):
Looks good. Left some comments but nothing blocking, apart from your comments on auto-approvals. Is that still an open discussion?

src/LiquidityPool.sol
ilinzweilin: nope, that would be an additional approval that we need to discuss

test/LiquidityPool.t.sol
ilinzweilin: yeah, agree. In this test it is about permission setup though. Other tests using this functions have balance checks

offerijns (approved):

offerijns (dismissed):

PR #87, opened 2023-08-28
Simplify LP, fix updating metadata, add permit support, and much more

offerijns:
- Add token metadata info to LPs
- Consolidate mapping based on poolId only
- Fix update token metadata
- Make poolId, trancheId, share and asset in LP immutable
- Pause admin separate access
- Remove determinism from LP factory
- Add request deposit / redeem with permit
- Add erc20 tests
- Re-enable admin tests
- Fix axelar router deployments


src/InvestmentManager.sol
```diff
@@ -54,14 +54,16 @@ interface AuthLike {
...
 }
 
-/// @dev centrifuge chain pool
+/// @dev Centrifuge pools
 struct Pool {
     uint64 poolId;
     uint256 createdAt;
     bool isActive;
+    mapping(bytes16 => Tranche) tranches;
+    mapping(address => bool) allowedCurrencies;
```
AStox, line 63: Love this!

src/LiquidityPool.sol
```diff
@@ -140,6 +144,13 @@ contract LiquidityPool is Auth {
         investmentManager.requestDeposit(assets, owner);
     }
 
+    function requestDepositWithPermit(uint256 assets, address owner, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
```
AStox, line 147: Maybe I'm just out of the loop, but it feels like this could use a `@dev` comment.

src/LiquidityPool.sol
```diff
@@ -170,6 +181,13 @@ contract LiquidityPool is Auth {
         investmentManager.requestRedeem(shares, owner);
     }
 
+    function requestRedeemWithPermit(uint256 shares, address owner, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
```
AStox, line 184: Likewise, it's not immediately obvious to me what v,r, and s should be, or where I can go to find that out. Since it's a public function, I think a quick explainer in a `@dev` comment would be nice

test/LiquidityPool.t.sol
```diff
@@ -214,6 +214,93 @@ contract LiquidityPoolTest is Test {
...
+                )
+            )
+        );
+
+        lPool.requestDepositWithPermit(amount, investor, block.timestamp, v, r, s);
+
+        // To avoid stack too deep errors
+        delete v;
+        delete r;
+        delete s;
```
AStox, line 260: Wow, I didn't know you could delete local variables!

AStox (approved):
Looks good. No blocking comments, but it would be nice to describe the `permit` a bit more in `@dev` comments.

src/LiquidityPool.sol
offerijns: Ticket created: https://github.com/centrifuge/liquidity-pools/issues/88

ilinzweilin (approved):

PR #89, opened 2023-08-29
Update relayer logic

offerijns:


src/gateway/routers/axelar/EVMRouter.sol
```diff
@@ -48,7 +48,7 @@ contract AxelarEVMRouter is Auth, AxelarExecutableLike {
     GatewayLike public gateway;
 
     string public constant axelarCentrifugeChainId = "Moonbeam";
-    string public constant axelarCentrifugeChainAddress = "0x7100fc671a443920606F01D5FD12dF72776a25CC";
+    string public constant axelarCentrifugeChainAddress = "0x56c4Db5bEaD29FC19158aA1f85673D9865732be4";
```
mustermeiszer, line 51: Why do we also change the EVM router? Shouldn't this only affect the relayer over moonbeam?

offerijns: Since the EVM router is sending the messages over Axelar to the relayer address as the destination

AStox (approved):
Looks good!

AStox (approved):

PR #94, opened 2023-08-30
Add trusted forwarder to TrancheToken

ilinzweilin:


src/token/IERC20.sol
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: MIT
```
offerijns, line 1: Can we rename this file to `ERC20Like.sol`?

Maybe we should actually create an interfaces folder. But let's decide how to organize that properly later

src/LiquidityPool.sol
```diff
@@ -193,7 +176,7 @@ contract LiquidityPool is Auth {
     function requestRedeemWithPermit(uint256 shares, address owner, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
         public
     {
-        share.permit(owner, address(this), shares, deadline, v, r, s);
+        share.permit(owner, address(investmentManager), shares, deadline, v, r, s);
```
offerijns, line 179: This makes me happy :)

src/InvestmentManager.sol
```diff
@@ -573,12 +576,14 @@ contract InvestmentManager is Auth {
...
 
-        EscrowLike(escrow).approve(tranche.token, liquidityPool, type(uint256).max);
         tranche.liquidityPools[_currency] = liquidityPool;
         wards[liquidityPool] = 1;
 
         // enable LP to take the liquidity pool tokens out of escrow in case if investments
         AuthLike(tranche.token).rely(liquidityPool); // add liquidityPool as ward on tranche Token
+        ERC2771Like(tranche.token).addTrustedForwarder(liquidityPool);
+        EscrowLike(escrow).approve(liquidityPool, address(this), type(uint256).max); // approve investment manager on tranche token for coordinating transfers
+        EscrowLike(escrow).approve(liquidityPool, liquidityPool, type(uint256).max); // approve liquidityPool on tranche token to be able to burn
```
offerijns, line 586: I need to better understand why this is the way it is, but will dive into that later, just leaving a note here as a reminder

src/LiquidityPool.sol
```diff
@@ -270,46 +253,40 @@ contract LiquidityPool is Auth {
...
-        returns (bool)
-    {
-        // discuss if we should add this here
-        // approveForOwner(sender, address(this), amount);
-        return share.transferFrom(sender, recipient, amount);
-    }
-
-    function transfer(address recipient, uint256 amount) public returns (bool) {
-        return transferFrom(msg.sender, recipient, amount);
+    function allowance(address owner, address spender) public view returns (uint256) {
```
offerijns, line 260: We have the generic ERC20 tests in `test/ERC20.t.sol`, we should adapt these to also be used for the liquidity pool to ensure it is compliant.

offerijns (approved):

PR #96, opened 2023-08-31
Update relayer payload

offerijns:


mustermeiszer (approved):
Looks good to me, but my Solidity is still jung and inexperienced!

PR #100, opened 2023-09-01
Add support for 128 byte token names

AStox:
Adds support for 128 byte token Names instead of only support 32byte token names and padding them to 128 bytes


src/gateway/Messages.sol
```diff
@@ -820,6 +817,38 @@ library Messages {
     }
 
     // TODO: should be moved to a util contract
+    function stringToBytes128(string memory source) internal pure returns (bytes memory) {
```
offerijns, line 820: I would move these and the bytes32 similar methods to a new `src/util/StringLib.sol`. We should also add some basic unit tests for these then.

src/gateway/Messages.sol
```diff
@@ -846,6 +840,38 @@ library Messages {
     }
 
     // TODO: should be moved to a util contract
```
offerijns, line 842:
```suggestion
```

src/gateway/Messages.sol
```diff
@@ -846,6 +840,38 @@ library Messages {
...
+    function stringToBytes128(string memory source) internal pure returns (bytes memory) {
```
offerijns, line 843: Let's adopt the _ pattern for internal functions
```suggestion
    function _stringToBytes128(string memory source) internal pure returns (bytes memory) {
```

src/gateway/Messages.sol
```diff
@@ -846,6 +840,38 @@ library Messages {
...
+                result[i] = temp[i];
+            } else {
+                result[i] = 0x00;
+            }
+        }
+
+        return result;
+    }
+
+    function bytes128ToString(bytes memory _bytes128) internal pure returns (string memory) {
```
offerijns, line 858:
```suggestion
    function _bytes128ToString(bytes memory _bytes128) internal pure returns (string memory) {
```

src/gateway/Messages.sol
```diff
@@ -846,6 +840,38 @@ library Messages {
...
+        bytes memory bytesArray = new bytes(i);
+
+        for (uint8 j = 0; j < i; j++) {
+            bytesArray[j] = _bytes128[j];
+        }
+
+        return string(bytesArray);
+    }
+
     function stringToBytes32(string memory source) internal pure returns (bytes32 result) {
```
offerijns, line 875:
```suggestion
    function _stringToBytes32(string memory source) internal pure returns (bytes32 result) {
```

src/gateway/Messages.sol
AStox: Want me to update the internal functions in TestSetup.t.sol too?

offerijns: Yes please do!

offerijns (approved):

PR #101, opened 2023-09-01
Add user escrow

offerijns:


src/InvestmentManager.sol
```diff
@@ -255,6 +244,9 @@ contract InvestmentManager is Auth {
         lpValues.maxWithdraw = lpValues.maxWithdraw + currencyPayout;
         lpValues.maxRedeem = lpValues.maxRedeem + trancheTokensPayout;
 
+        escrow.approve(_currency, address(userEscrow), currencyPayout);
```
ilinzweilin, line 247: I think we can make that a general approval when we add a new supported currency

src/UserEscrow.sol
```diff
@@ -0,0 +1,41 @@
...
+    mapping(address => mapping(address => uint256)) destinations; // map by token and destination
+
+    constructor() {
+        wards[msg.sender] = 1;
+        emit Rely(msg.sender);
+    }
+
+    // --- Token approvals ---
+    function transferIn(address token, address source, address destination, uint256 amount) external auth {
+        require(TransferLike(token).transferFrom(source, address(this), amount), "UserEscrow/transfer-failed");
```
ilinzweilin, line 29: if we also want to avoid transferFrom here, we need to call an auth transferOut function on the escrow

ilinzweilin (changes requested):

ilinzweilin (approved):

PR #106, opened 2023-09-02
Update relayer, add decrease methods

offerijns:
- Update relayer logic, remove inheritance
- Remove inheritance from Axelar EVM router
- Add decrease* methods
- Pass wards as arrays to factories for more flexibility


src/LiquidityPool.sol
```diff
@@ -245,15 +247,23 @@ contract LiquidityPool is Auth, ERC20Like {
...
         emit DepositCollected(receiver);
     }
 
     function collectRedeem(address receiver) public {
-        investmentManager.collectRedeem(poolId, trancheId, receiver, asset);
+        investmentManager.collectRedeem(receiver);
         emit RedeemCollected(receiver);
     }
 
+    function decreaseDepositRequest(uint256 assets) public {
```
ilinzweilin, line 259: you can use the ´withCurrencyApproval´ modifier here

src/LiquidityPool.sol
```diff
@@ -245,15 +247,23 @@ contract LiquidityPool is Auth, ERC20Like {
...
+        investmentManager.decreaseDepositRequest(assets, msg.sender);
+    }
+
+    function decreaseRedeemRequest(uint256 shares) public {
```
ilinzweilin, line 263: you can use the ´withTokenApproval´ modifier here

src/PoolManager.sol
```diff
@@ -314,8 +325,10 @@ contract PoolManager is Auth {
         require(liquidityPool == address(0), "PoolManager/liquidityPool-already-deployed");
         require(pools[poolId].createdAt != 0, "PoolManager/pool-does-not-exist");
 
+        address[] memory liquidityPoolWards = new address[](1);
+        liquidityPoolWards[0] = address(investmentManager);
```
ilinzweilin, line 329: why passing the wards to the factory, if we pass the investment manager address anyway.

ilinzweilin (changes requested):

src/PoolManager.sol
offerijns: So that in the future if we might need another ward, we don't need to modify the factory contract, which would be a bytecode change and thus change the address.

ilinzweilin (approved):

PR #113, opened 2023-09-05
Adapt LP price calculations

offerijns:


src/InvestmentManager.sol
```diff
@@ -317,7 +319,33 @@ contract InvestmentManager is Auth {
...
+        (uint8 currencyDecimals, uint8 trancheTokenDecimals) = _getPoolDecimals(liquidityPool);
+
+        shares = assets.mulDiv(
+            10 ** (PRICE_DECIMALS + trancheTokenDecimals - currencyDecimals),
+            LiquidityPoolLike(liquidityPool).latestPrice(),
+            MathLib.Rounding.Down
+        );
+    }
+
+    /// @dev Calculates the asset value for an amount of shares / tranche tokens provided. The calcultion is based on the token price from the most recent epoch retrieved from Centrifuge chain.
```
AStox, line 337:
```suggestion
    /// @dev Calculates the asset value for an amount of shares / tranche tokens provided. The calculation is based on the token price from the most recent epoch retrieved from Centrifuge chain.
```

src/InvestmentManager.sol
```diff
@@ -317,7 +319,33 @@ contract InvestmentManager is Auth {
     }
 
     // --- View functions ---
+    function totalAssets(uint256 totalSupply, address liquidityPool) public view returns (uint256 totalAssets) {
+        totalAssets = convertToAssets(totalSupply, liquidityPool);
+    }
+
+    /// @dev Calculates the amount of shares / tranche tokens that any user would get for the amount of assets provided. The calcultion is based on the token price from the most recent epoch retrieved from Centrifuge chain.
```
AStox, line 326:
```suggestion
    /// @dev Calculates the amount of shares / tranche tokens that any user would get for the amount of assets provided. The calculation is based on the token price from the most recent epoch retrieved from Centrifuge chain.
```

src/InvestmentManager.sol
```diff
@@ -356,32 +384,33 @@ contract InvestmentManager is Auth {
...
 
     /// @return currencyAmount is type of uin256 to support the EIP4626 Liquidity Pool interface
-    function previewWithdraw(address user, address liquidityPool, uint256 trancheTokenAmount)
+    function previewRedeem(address user, address liquidityPool, uint256 trancheTokenAmount)
         public
         view
         returns (uint256 currencyAmount)
     {
-        uint128 redeemPrice = calculateRedeemPrice(user, liquidityPool);
+        require(trancheTokenAmount < type(uint128).max, "InvestmentManager/amount-too-large");
```
AStox, line 412: Should we also add a similar check to `previewWithdraw()`?

src/InvestmentManager.sol
```diff
@@ -635,36 +661,34 @@ contract InvestmentManager is Auth {
...
+    function _toPriceDecimals(uint128 _value, uint8 decimals, address liquidityPool)
         internal
         view
-        returns (uint128 value)
+        returns (uint256 value)
     {
-        (uint8 maxDecimals,,) = _getPoolDecimals(liquidityPool);
-        if (maxDecimals == decimals) return _value;
-        return _toUint128(_value * 10 ** (maxDecimals - decimals));
+        if (PRICE_DECIMALS == decimals) return uint256(_value);
```
AStox, line 671: This line is unneeded isn't it? `_value * 10 ** 0` returns `_value` anyway. Or is there some weird solidity edge-case here I'm not seeing?

src/InvestmentManager.sol
```diff
@@ -512,75 +541,72 @@ contract InvestmentManager is Auth {
...
+        view
+        returns (uint256 depositPrice)
+    {
+        (uint8 currencyDecimals, uint8 trancheTokenDecimals) = _getPoolDecimals(liquidityPool);
+        uint256 currencyAmountInPriceDecimals = _toPriceDecimals(currencyAmount, currencyDecimals, liquidityPool);
+        uint256 trancheTokenAmountInPriceDecimals =
+            _toPriceDecimals(trancheTokenAmount, trancheTokenDecimals, liquidityPool);
+
+        depositPrice = currencyAmountInPriceDecimals.mulDiv(
+            10 ** PRICE_DECIMALS, trancheTokenAmountInPriceDecimals, MathLib.Rounding.Down
```
AStox, line 573: I think this still ends up being the same calculation in the end, right? Just presented in a more digestible way.

AStox (commented):
I'm not sure I follow the math in your tests. Still need to go over it, but I'll leave some other comments for you in the meantime.

src/InvestmentManager.sol
offerijns: This was a leftover from some bugfixing, should have been removed.

src/InvestmentManager.sol
offerijns: It just is more efficient to skip the pow multiplication below if we don't need it.

src/InvestmentManager.sol
offerijns: Yes indeed, just refactored a bit.

ilinzweilin (dismissed):

ilinzweilin (approved):

PR #123, opened 2023-09-06
increase test coverage

ilinzweilin:


src/UserEscrow.sol
```diff
@@ -35,7 +35,14 @@ contract UserEscrow is Auth {
...
         require(
-            receiver == destination || (ERC20Like(token).allowance(destination, receiver) >= amount),
+            /**
+             * @dev
+             * transferOut can only be initiated by the destination address or an authorized admin.
+             * The check is just an additional protection to secure destination funds in case of compromized auth.
+             * Since userEscrow is not able to decrease the allowance for the receiver,
+             * a transfer is only possible in case receiver has received the full allowance from destination address.
+             */
+            receiver == destination || (ERC20Like(token).allowance(destination, receiver) == type(uint256).max),
```
offerijns, line 45: To be consistent
```suggestion
             /// @dev transferOut can only be initiated by the destination address or an authorized admin.
             ///      The check is just an additional protection to secure destination funds in case of compromized auth.
             ///      Since userEscrow is not able to decrease the allowance for the receiver,
             ///      a transfer is only possible in case receiver has received the full allowance from destination address.
            receiver == destination || (ERC20Like(token).allowance(destination, receiver) == type(uint256).max),
```

test/LiquidityPool.t.sol
```diff
@@ -96,6 +96,66 @@ contract LiquidityPoolTest is TestSetup {
         investor.requestDeposit(lPool_, amount / 2, address(investor));
     }
 
+    // --- uint128 type checks ---
+    // Make sure all function calls would fail when overflow uint128
+    function testAssertUint128(
```
offerijns, line 101: Nice!

PR #153, opened 2023-09-21
Migration contracts & migration tests

AStox:


test/migrations/MigratedLiquidityPool.t.sol
```diff
@@ -0,0 +1,63 @@
...
+        // clean up
+        newLiquidityPool.deny(address(this));
+        root.denyContract(address(investmentManager), address(this));
+        root.denyContract(address(escrow), address(this));
+        root.deny(address(this));
+
+        // verify permissions
+        verifyLiquidityPoolPermissions(LiquidityPool(_lPool), newLiquidityPool);
+
+        // TODO: test that everything is working
```
hieronx, line 42: This should fail as `addLiquidityPool` needs to be called on the tranche token.

Also, we should call `removeLiquidityPool` for the old one, and make sure to also do
`poolManager.escrow().approve(address(trancheToken), oldLiquidityPool, 0);`

PR #163, opened 2023-09-30
Sanity check executions

hieronx:


src/InvestmentManager.sol
```diff
@@ -343,12 +343,12 @@ contract InvestmentManager is Auth {
         lpValues.maxMint = lpValues.maxMint + trancheTokensPayout;
         lpValues.remainingInvestOrder = remainingInvestOrder;
 
+        LiquidityPoolLike(liquidityPool).updatePrice(_toUint128(lpValues.depositPrice));
```
ilinzweilin, line 346: I figured we should update with the deposit price here. As it is averaged for the user & is influenced by prev epochs. For the update, we would need to calculate a price only based on values from the executed order

src/InvestmentManager.sol
```diff
@@ -377,14 +382,14 @@ contract InvestmentManager is Auth {
         lpValues.maxWithdraw = lpValues.maxWithdraw + currencyPayout;
         lpValues.remainingRedeemOrder = remainingRedeemOrder;
 
+        LiquidityPoolLike(liquidityPool).updatePrice(_toUint128(lpValues.redeemPrice));
```
ilinzweilin, line 385: same here. regarding the price

src/InvestmentManager.sol
```diff
@@ -365,8 +365,13 @@ contract InvestmentManager is Auth {
         address _currency = poolManager.currencyIdToAddress(currency);
         address liquidityPool = poolManager.getLiquidityPool(poolId, trancheId, _currency);
         require(liquidityPool != address(0), "InvestmentManager/tranche-does-not-exist");
-
         LPValues storage lpValues = orderbook[liquidityPool][recipient];
+
+        // Sanity check that there is indeed an outstanding order for this recipient
+        // before transferring funds to the user escrow
+        require(lpValues.remainingRedeemOrder != 0, "InvestmentManager/no-outstanding-order");
```
ilinzweilin, line 372: here we might get an issue, if the tx order is not sequential.
Example:
- order is fullfilled within 2 epochs:
- the second tx from the second epoch comes in first & sets the remaining order to 0.
-  when the tx from the first epoch execution arrives - it fails, since remainingRedeemOrder was already set to 0 by the second epoch tx.

src/InvestmentManager.sol
```diff
@@ -402,20 +407,24 @@ contract InvestmentManager is Auth {
         address liquidityPool = poolManager.getLiquidityPool(poolId, trancheId, _currency);
         require(liquidityPool != address(0), "InvestmentManager/tranche-does-not-exist");
         require(_currency == LiquidityPoolLike(liquidityPool).asset(), "InvestmentManager/not-tranche-currency");
+        LPValues storage lpValues = orderbook[liquidityPool][user];
 
-        // Transfer currency amount to userEscrow
-        userEscrow.transferIn(_currency, address(escrow), user, currencyPayout);
+        // Sanity check that there is indeed an outstanding order for this recipient
+        // before transferring funds to the user escrow
+        require(lpValues.remainingInvestOrder != 0, "InvestmentManager/no-outstanding-order");
```
ilinzweilin, line 414: same here regarding tx order

ilinzweilin (changes requested):

src/InvestmentManager.sol
AStox: This scenario would be a good test case

ilinzweilin (approved):

PR #176, opened 2023-10-11
Add C4 and SRLabs audit reports

hieronx:


lemunozm (approved):
Approved (EDIT: but not permissions enough in this repo 😅)

Although not sure if we want to add a pdf to the codebase as it counts as binaries and increases the size of the repo significantly. Is there any way to generate it or link it from a Google Drive?

AStox (approved):

PR #178, opened 2023-10-13
Align approval flow and interface with EIP draft

hieronx:


src/InvestmentManager.sol
```diff
@@ -489,11 +506,15 @@ contract InvestmentManager is Auth {
         currencyAmount = uint256(_calculateCurrencyAmount(trancheTokenAmount, liquidityPool, state.redeemPrice));
     }
 
-    function userDepositRequest(address liquidityPool, address user) public view returns (uint256 currencyAmount) {
+    function pendingDepositRequest(address liquidityPool, address user) public view returns (uint256 currencyAmount) {
```
AStox, line 509: nice name change 👍

src/InvestmentManager.sol
```diff
@@ -189,7 +192,11 @@ contract InvestmentManager is Auth {
     ///         in case their orders got fulfilled.
     /// @dev    The user tranche tokens required to fulfill the redemption request have to be locked,
     ///         even though the currency payout can only happen after epoch execution.
-    function requestRedeem(address liquidityPool, uint256 trancheTokenAmount, address user) public auth {
+    function requestRedeem(address liquidityPool, uint256 trancheTokenAmount, address user)
```
AStox, line 195: like `requestDeposit`, I think renaming `user` to `operator` is more descriptive and consistent

src/LiquidityPool.sol
```diff
@@ -300,19 +312,23 @@ contract LiquidityPool is Auth, IERC4626 {
         return share.allowance(owner, spender);
     }
 
-    function transferFrom(address, address, uint256) public returns (bool) {
-        (bool success, bytes memory data) = address(share).call(bytes.concat(msg.data, bytes20(msg.sender)));
+    function transferFrom(address from, address to, uint256 value) public returns (bool) {
+        (bool success, bytes memory data) = address(share).call(
+            bytes.concat(
+                abi.encodeWithSignature("transferFrom(address,address,uint256)", from, to, value), bytes20(msg.sender)
```
AStox, line 318: whats the difference between this and the previous call?

AStox: I like that its more explicit now, but curious what the reasoning was here.

src/InvestmentManager.sol
```diff
@@ -153,33 +153,36 @@ contract InvestmentManager is Auth {
...
 
         uint64 poolId = lPool.poolId();
         address currency = lPool.asset();
         require(poolManager.isAllowedAsInvestmentCurrency(poolId, currency), "InvestmentManager/currency-not-allowed");
+
+        require(
+            _checkTransferRestriction(liquidityPool, address(0), sender, 0), "InvestmentManager/sender-is-restricted"
+        );
         require(
             _checkTransferRestriction(liquidityPool, address(0), user, convertToShares(liquidityPool, currencyAmount)),
```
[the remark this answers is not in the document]
hieronx: Every account interacting with the contracts needs to be on the memberlist, so also an integrating party yeah.

src/LiquidityPool.sol
hieronx: This gets quite deep, but is definitely good to fully understand.

Previously, `transferFrom` was only called externally. Now it is called externally, but also used internally in `requestRedeem`.

Before it used `msg.data + msg.sender` as the full payload. `msg.data` is basically the incoming encoded call. So if you externally call `transferFrom`, it will be `encode(transferFrom(from, to, value))`. But, if you call it from `requestRedeem`, the `msg.data` here will also be `requestRedeem(shares)`. So the call to the tranche token will actually have the request redeem signature, and therefore of course fail.

So instead we are now manually doing the ABI encoding here, rather than using `msg.data`, so it works both for external and internal calls to this function.

src/LiquidityPool.sol
```diff
@@ -215,23 +210,36 @@ contract LiquidityPool is Auth, IERC4626 {
...
+        uint256 assets,
+        address operator,
+        address owner,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) public {
+        _withPermit(asset, owner, address(this), assets, deadline, v, r, s);
+        require(manager.requestDeposit(address(this), assets, owner, operator), "LiquidityPool/request-deposit-failed");
```
ilinzweilin, line 233: I don't think that we can pass the operator here.
Malicious operator that is also on the restricted token allowlist could frontrun the permit and pass their own operator address to the function

src/LiquidityPool.sol
```diff
@@ -249,11 +257,13 @@ contract LiquidityPool is Auth, IERC4626 {
...
-    /// @notice Request can only be called by the owner of the shares
+    ///         DOES support flow where owner != msg.sender but has allowance to spend its shares
     ///         Shares are locked in the escrow on request submission
-    function requestRedeem(uint256 shares) public {
-        manager.requestRedeem(address(this), shares, msg.sender);
-        emit RedeemRequest(msg.sender, shares);
+    function requestRedeem(uint256 shares, address operator, address owner) public {
+        require(share.balanceOf(owner) >= shares, "LiquidityPool/insufficient-balance");
+        require(manager.requestRedeem(address(this), shares, operator), "LiquidityPool/request-redeem-failed");
+        require(transferFrom(owner, address(escrow), shares), "LiquidityPool/transfer-failed");
```
ilinzweilin, line 265: we need to add a comment here that this is only possible because of the trusted forwarder pattern -> msg.sender is forwarded and the call can only be executed, if msg.sender has owner's approval to spent tokens

ilinzweilin (changes requested):

src/InvestmentManager.sol
```diff
@@ -153,33 +153,36 @@ contract InvestmentManager is Auth {
     ///         proceed with tranche token payouts in case their orders got fulfilled.
     /// @dev    The user currency amount required to fulfill the deposit request have to be locked,
     ///         even though the tranche token payout can only happen after epoch execution.
-    function requestDeposit(address liquidityPool, uint256 currencyAmount, address user) public auth {
+    function requestDeposit(address liquidityPool, uint256 currencyAmount, address sender, address user)
```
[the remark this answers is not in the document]
hieronx: This contract is anyways mapping from the liquidity pool naming to the centrifuge naming (also asset => currency, share => tranche token, etc). I think using `user` consistently in this contract makes more sense.

ilinzweilin (approved):

PR #193, opened 2023-10-24
Check transfer restriction on request redeem

hieronx:


src/InvestmentManager.sol
```diff
@@ -207,6 +207,12 @@ contract InvestmentManager is Auth {
             "InvestmentManager/currency-not-allowed"
         );
 
+        // Check that the user (operator) is a non-frozen member
+        require(
+            _checkTransferRestriction(liquidityPool, user, user, convertToAssets(liquidityPool, trancheTokenAmount)),
```
ilinzweilin, line 212: i think to should be address(escrow). Otherwise users that are unfrozen not on the allowlist won't able to redeem

AStox (approved):

PR #197, opened 2023-10-24
Revert cancel rely if no target is scheduled

hieronx:


ilinzweilin (approved):

src/Root.sol
```diff
@@ -74,6 +74,7 @@ contract Root is Auth {
 
     /// @notice Cancel a pending scheduled rely
     function cancelRely(address target) external auth {
+        require(schedule[target] != 0, "Root/target-not-scheduled");
```
ilinzweilin, line 77: shouldn't it be `address(0`)

hieronx: The `schedule[target]` value is the timestamp when it can be relied

ilinzweilin: aaaaah, nvm 🙈

PR #199, opened 2023-10-26
Clean up events, update liquidity pool docs

hieronx:


src/PoolManager.sol
```diff
@@ -442,31 +440,21 @@ contract PoolManager is Auth {
...
     {
-        TrancheTokenPrice memory value = pools[poolId].tranches[trancheId].prices[currencyAddress];
+        TrancheTokenPrice memory value = pools[poolId].tranches[trancheId].prices[currency];
         price = value.price;
         computedAt = value.computedAt;
     }
 
-    function isAllowedAsInvestmentCurrency(uint64 poolId, address currencyAddress) public view returns (bool) {
-        uint128 currency = currencyAddressToId[currencyAddress];
-        if (currency == 0) {
```
hieronx, line 465: This is already checked in `addCurrency`. And `allowInvestmentCurrency` can only be called after `addCurrency`.

NunoAlexandre (approved):
The changes look good to me. One question tho: is it an idea that we verify the emitted events in the tests? Not sure if that's an idiomatic thing to do in Solidity but I'd see value in having that safe-guard in place. wdyt?

hieronx:
> One question tho: is it an idea that we verify the emitted events in the tests? Not sure if that's an idiomatic thing to do in Solidity but I'd see value in having that safe-guard in place. wdyt?

Yeah we definitely should add tests for this! We just haven't gotten to this yet. We can use https://book.getfoundry.sh/cheatcodes/expect-emit for this

NunoAlexandre:
> > One question tho: is it an idea that we verify the emitted events in the tests? Not sure if that's an idiomatic thing to do in Solidity but I'd see value in having that safe-guard in place. wdyt?
> 
> Yeah we definitely should add tests for this! We just haven't gotten to this yet. We can use https://book.getfoundry.sh/cheatcodes/expect-emit for this

Nice, created an issue for that: https://github.com/centrifuge/liquidity-pools/issues/201

ilinzweilin (approved):

AStox (approved):
Looks good!

PR #202, opened 2023-10-30
Emit claimable events from liquidity pool

hieronx:


src/InvestmentManager.sol
```diff
@@ -74,9 +76,9 @@ struct InvestmentState {
     /// @dev Weighted average price of redemptions, used to convert maxWithdraw to maxRedeem
     uint256 redeemPrice;
     /// @dev Remaining invest (deposit) order in currency
-    uint128 remainingDepositRequest;
+    uint128 pendingDepositRequest;
```
NunoAlexandre, line 79: Now the docs and the variable names don't match. I.e, the doc string calls it "remaining" and the code variable name "pending". We should update the doc.

NunoAlexandre (commented):
One thing to fix. For my understanding, the idea here is to have all the relevant events being emitted by the same contract instead of having them across them, to facilitate indexing and monitoring?

src/InvestmentManager.sol
hieronx: I think that's a good thing because it explains the variable in other words. If the comment is just the same as the variable, we might as well remove the comment.

hieronx: Also the naming in this contract is bridging between the liquidity pool (eip7540 standard naming of `pendingDepositRequest` and the naming we use in our messages (` remainingInvestOrder`), this comment makes that link clear. Similar to how `maxWithdraw` in the comment is described as the currency

hieronx:
> For my understanding, the idea here is to have all the relevant events being emitted by the same contract instead of having them across them, to facilitate indexing and monitoring?

Yes exactly!

src/InvestmentManager.sol
NunoAlexandre: > I think that's a good thing because it explains the variable in other words. 

I have to disagree. I understand what you are saying but imagine that we would go on using synonyms randomly around the code.

> If the comment is just the same as the variable, we might as well remove the comment.

It's not a worse alternative in this case imho. Or, even better, we can actually document what this value means: where it is based off, how it changes, etc.

Not a blocker for me 👍

NunoAlexandre (approved):

src/InvestmentManager.sol
hieronx: We are forced to use synonyms already. We have to use the `pendingDepositRequest` naming because it's in the EIP7540 standard, and we are focused to use `remainingInvestOrder` as that is now included in the messages we defined cross-chain

ilinzweilin (approved):

PR #206, opened 2023-11-14
Support ERC-7540 changes and ERC-7575

hieronx:


src/LiquidityPool.sol
```diff
@@ -93,13 +103,44 @@ contract LiquidityPool is Auth, IERC7540 {
...
+    {
+        require(owner == msg.sender, "LiquidityPool/not-msg-sender");
+        require(IERC20(asset).balanceOf(owner) >= assets, "LiquidityPool/insufficient-balance");
+
+        require(manager.requestDeposit(address(this), assets, receiver, owner), "LiquidityPool/request-deposit-failed");
+        SafeTransferLib.safeTransferFrom(asset, owner, address(escrow), assets);
+
         require(
-            manager.requestDeposit(address(this), assets, receiver, msg.sender), "LiquidityPool/request-deposit-failed"
+            data.length == 0 || receiver.code.length == 0
```
hieronx, line 117: The callback is only called if nonzero data was passed, and the receiver is a contract.

There should not be any reentrancy issues because the callback occurs after all state changes (only the event emission and the return call happen after this)

NunoAlexandre (dismissed):
To the best of my humble ability here, I don't see any issues and generally the changes are very sound 👌

test/LiquidityPool.t.sol
```diff
@@ -93,8 +95,8 @@ contract LiquidityPoolTest is TestSetup {
     // --- erc165 checks ---
     function testERC165Support(bytes4 unsupportedInterfaceId) public {
         bytes4 erc165 = 0x01ffc9a7;
-        bytes4 erc7540Deposit = 0xea446681;
-        bytes4 erc7540Redeem = 0x2e9dd5bd;
+        bytes4 erc7540Deposit = 0x1683f250;
+        bytes4 erc7540Redeem = 0x0899cb0b;
```
AStox, line 99: Just curious. What are these hex representations of?

AStox (approved):

test/LiquidityPool.t.sol
hieronx: They are the interface IDs. See https://eips.ethereum.org/EIPS/eip-165

src/LiquidityPool.sol
```diff
@@ -320,38 +320,7 @@ contract LiquidityPool is Auth, IERC7540 {
...
-    function balanceOf(address owner) external view returns (uint256) {
-        return share.balanceOf(owner);
-    }
-
-    /// @inheritdoc IERC20
-    function allowance(address owner, address spender) external view returns (uint256) {
-        return share.allowance(owner, spender);
-    }
-
-    /// @inheritdoc IERC20
```
NunoAlexandre, line 287: nice cleanup 👌

AStox (approved):
Looks good!

ilinzweilin (approved):

PR #207, opened 2023-11-14
Add Axelar forwarder contract

mustermeiszer:
Adds a `Forwarder` contract to the codebase. 

* `Forwarder` implements `IAxelarExecutable` (`AxelarGatewayLike` in our codebase) allowing it to be used as a destination of messages going over the Axelar bridge
* `Forwarder` checks whether messages are validated by the Axelar bridge
* `Forwarder` forwards always do the rust-based precompile

Reasons: 
The precompile is currently just processing all messages that come from a gateway contract address. As it is hard to check the validity of messages by querying the gateways storage, we use the `Forwarder` to do that for the precompile and we use the `Forwarder` as our allowed source to submit to the precompile.


mustermeiszer:
@ilinzweilin @hieronx this should be good for a review

src/gateway/routers/axelar/LocalGateway.sol
```diff
@@ -0,0 +1,29 @@
...
+        bytes32 commandId,
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes32 payloadHash
+    ) external returns (bool);
+}
+
+// A contract to be deployed on Centrifuge-EVM in order to forward axelar tx to
+// the precompile.
+contract PassthroughGateway {
```
hieronx, line 15: I don't understand the usage of this contract 🤔

src/gateway/routers/axelar/Router.sol
```diff
@@ -85,6 +87,8 @@ contract AxelarRouter is Auth {
 
     // --- Outgoing ---
     function send(bytes calldata message) public onlyGateway {
-        axelarGateway.callContract(CENTRIFUGE_CHAIN_ID, LP_PRECOMPILE_ADDRESS, message);
+        axelarGateway.callContract(
+            CENTRIFUGE_CHAIN_ID, BytesLib.toHex(abi.encodePacked(centrifugeAxelarExecutable)), message
```
hieronx, line 91: Worth noting that this is making every single call more expensive, by adding the abi encoding and hex conversion. I'm not 100% sure that's worth it vs just continuing to keep it hardcoded?

src/util/BytesLib.sol
```diff
@@ -130,4 +130,18 @@ library BytesLib {
 
         return tempBytes16;
     }
+
+    function toHex(bytes memory _bytes) public pure returns (string memory) {
```
hieronx, line 134: Any source for this?

src/gateway/routers/axelar/Router.sol
mustermeiszer: Did not know that. Will hardcode

src/util/BytesLib.sol
mustermeiszer: Just this one: https://ethereum.stackexchange.com/questions/126899/convert-bytes-to-hexadecimal-string-in-solidity

mustermeiszer: I can add tests if you show me how.

src/gateway/routers/axelar/LocalGateway.sol
hieronx: But isn't that the `Forwarder`? I guess I'm just not following the difference

src/gateway/routers/axelar/Router.sol
```diff
@@ -57,9 +59,11 @@ contract AxelarRouter is Auth {
     }
 
     // --- Administration ---
-    function file(bytes32 what, address data) external auth {
+    function file(bytes32 what, bytes memory data) external auth {
```
ilinzweilin, line 62: you can overload the file function and support both cases `address` and `bytes memory`. That way you can safe the `BytesLib` import

src/gateway/routers/axelar/Router.sol
```diff
@@ -23,18 +23,20 @@ interface GatewayLike {
 /// @notice Routing contract that integrates with an Axelar Gateway
 contract AxelarRouter is Auth {
     string internal constant CENTRIFUGE_CHAIN_ID = "centrifuge";
-    string internal constant CENTRIFUGE_CHAIN_ADDRESS = "0x7369626cef070000000000000000000000000000";
-    string internal constant LP_PRECOMPILE_ADDRESS = "0x0000000000000000000000000000000000002048";
+    string internal constant CENTRIFUGE_CHAIN_ADDRESS = "0x7369626CEF070000000000000000000000000000";
```
hieronx, line 26: Just confirming: this definitely should be uppercased now?

hieronx (dismissed):

hieronx (approved):

PR #209, opened 2023-11-16
setup flex restrictions

ilinzweilin:


src/PoolManager.sol
```diff
@@ -440,6 +442,10 @@ contract PoolManager is Auth {
         return tranche.token;
     }
 
+    function getTrancheTokenRestriction(uint64 poolId, bytes16 trancheId) public view returns (uint8 restrictionSet) {
```
hieronx, line 445: I wouldn't add this method just for tests. It's also a bit confusing because the name implies it always gets the restriction, but it actually only works for undeployed tranches

src/token/Tranche.sol
```diff
@@ -79,7 +79,7 @@ contract TrancheToken is ERC20 {
         returns (bool success)
     {
         success = super.transferFrom(from, to, value);
-        if (success) restrictionManager.afterTransfer(_msgSender(), to, value);
+        if (success) restrictionManager.afterTransfer(from, to, value);
```
hieronx, line 82: Good catch!

src/util/Factory.sol
```diff
@@ -124,16 +124,17 @@ contract RestrictionManagerFactory is Auth {
         emit Rely(msg.sender);
     }
 
-    function newRestrictionManager(
-        uint8, /* restrictionSet */
-        address token,
-        address[] calldata restrictionManagerWards
-    ) public auth returns (address) {
+    function newRestrictionManager(uint8 restrictionSet, address token, address[] calldata restrictionManagerWards)
```
hieronx, line 127: `restrictionSet` is an unused param for now right? That's why I commented the name

test/gateway/Messages.t.sol
```diff
@@ -61,46 +61,52 @@ contract MessagesTest is Test {
         assertEq(decodedCurrency, uint256(currency));
     }
 
-    function testAddTranche() public {
+    function testAddTrancheMoin() public {
```
hieronx, line 64: I don't understand this name :)

test/gateway/Messages.t.sol
```diff
@@ -61,46 +61,52 @@ contract MessagesTest is Test {
...
         uint64 poolId = 1;
         bytes16 trancheId = bytes16(hex"811acd5b3f17c06841c7e41e9e04cb1b");
         string memory name = "Some Name";
         string memory symbol = "SYMBOL";
         uint8 decimals = 15;
+        uint8 restrictionSet = 2;
         bytes memory expectedHex =
-            hex"040000000000000001811acd5b3f17c06841c7e41e9e04cb1b536f6d65204e616d65000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000053594d424f4c00000000000000000000000000000000000000000000000000000f";
+            hex"040000000000000001811acd5b3f17c06841c7e41e9e04cb1b536f6d65204e616d65000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000053594d424f4c00000000000000000000000000000000000000000000000000000f02";
```
hieronx, line 72: We should change the message decoding to support also messages where the restriction set is not included, so the old and new version are supported, for backwards compatibility (otherwise we depend on centrifuge chain upgrading before release).

I think this can be quite easy by on this line: https://github.com/centrifuge/liquidity-pools/blob/0f77a1a008e082494b131e470bce022fcc86e0c8/src/gateway/Messages.sol#L187 Making it something like
```solidity
if (_msg.length > 186) restrictionSet = BytesLib.toUint8(_msg, 186);
```

hieronx (approved):

PR #214, opened 2023-11-22
Remove operator usage

hieronx:


NunoAlexandre (dismissed):

src/InvestmentManager.sol
```diff
@@ -135,7 +135,7 @@ contract InvestmentManager is Auth {
     ///         proceed with tranche token payouts in case their orders got fulfilled.
     /// @dev    The user currency amount required to fulfill the deposit request have to be locked,
     ///         even though the tranche token payout can only happen after epoch execution.
-    function requestDeposit(address liquidityPool, uint256 currencyAmount, address sender, address operator)
+    function requestDeposit(address liquidityPool, uint256 currencyAmount, address receiver, address owner)
```
AStox, line 138: I think these are backwards
```suggestion
    function requestDeposit(address liquidityPool, uint256 currencyAmount, address owner, address receiver)
```

ilinzweilin: should be fine @AStox as the LiquidityPool is calling this function in the following way: 
```
manager.requestDeposit(address(this), assets, receiver, msg.sender), "LiquidityPool/request-deposit-failed"
        );
```

src/LiquidityPool.sol
```diff
@@ -192,8 +189,8 @@ contract LiquidityPool is Auth, IERC7540 {
     }
 
     /// @inheritdoc IERC4626
-    function maxDeposit(address operator) external view returns (uint256 maxAssets) {
-        maxAssets = manager.maxDeposit(address(this), operator);
+    function maxDeposit(address owner) external view returns (uint256 maxAssets) {
```
ilinzweilin, line 192: I think `owner` is very confusing here, as it is not necessarily the initial owner that locked the tokens. It can also be a different receiver address... I would go with a more neutral naming like `user` and only distinguish between owner and receiver if required by the context.
Same applies to all the other `max` functions

src/LiquidityPool.sol
```diff
@@ -96,43 +96,40 @@ contract LiquidityPool is Auth, IERC7540 {
...
-        );
-        SafeTransferLib.safeTransferFrom(asset, msg.sender, address(escrow), assets);
-        emit DepositRequest(msg.sender, msg.sender, assets);
+        requestDeposit(assets, receiver);
     }
 
     /// @inheritdoc IERC7540Deposit
-    function pendingDepositRequest(address operator) external view returns (uint256 assets) {
-        assets = manager.pendingDepositRequest(address(this), operator);
+    function pendingDepositRequest(address owner) external view returns (uint256 assets) {
```
ilinzweilin, line 118: would also go with `user` here. Same reasoning as for the max - functions

src/LiquidityPool.sol
```diff
@@ -96,43 +96,40 @@ contract LiquidityPool is Auth, IERC7540 {
...
+        require(manager.requestRedeem(address(this), shares, receiver, owner), "LiquidityPool/request-redeem-failed");
         require(transferFrom(owner, address(escrow), shares), "LiquidityPool/transfer-failed");
-        emit RedeemRequest(msg.sender, operator, owner, shares);
+        emit RedeemRequest(msg.sender, receiver, owner, shares);
     }
 
     /// @inheritdoc IERC7540Redeem
-    function pendingRedeemRequest(address operator) external view returns (uint256 shares) {
-        shares = manager.pendingRedeemRequest(address(this), operator);
+    function pendingRedeemRequest(address owner) external view returns (uint256 shares) {
```
ilinzweilin, line 131: would also go with user here. Same reasoning as for the max - functions

src/LiquidityPool.sol
```diff
@@ -214,29 +211,29 @@ contract LiquidityPool is Auth, IERC7540 {
...
 
     /// @inheritdoc IERC4626
-    /// @notice DOES NOT support operator != msg.sender since shares are already transferred on requestRedeem
-    function withdraw(uint256 assets, address receiver, address operator) external returns (uint256 shares) {
-        require((msg.sender == operator), "LiquidityPool/not-the-operator");
-        shares = manager.withdraw(address(this), assets, receiver, operator);
-        emit Withdraw(msg.sender, receiver, operator, assets, shares);
+    /// @notice DOES NOT support owner != msg.sender since shares are already transferred on requestRedeem
+    function withdraw(uint256 assets, address receiver, address owner) external returns (uint256 shares) {
+        require((msg.sender == owner), "LiquidityPool/not-the-owner");
```
ilinzweilin, line 221: I think, here we need to emphasise that `owner` is very context dependent. The `owner` in this function is not necessarily the owner of the shares locked in escrow (passed in `requestRedeem`), but the receiver of the fullfilled order

src/InvestmentManager.sol
```diff
@@ -36,8 +36,8 @@ interface LiquidityPoolLike is ERC20Like {
     function trancheId() external view returns (bytes16);
     function asset() external view returns (address);
     function share() external view returns (address);
-    function emitDepositClaimable(address operator, uint256 assets, uint256 shares) external;
-    function emitRedeemClaimable(address operator, uint256 assets, uint256 shares) external;
+    function emitDepositClaimable(address owner, uint256 assets, uint256 shares) external;
```
ilinzweilin, line 39: here I would also go with `user` same reasoning as with the max -functions

ilinzweilin (changes requested):

src/LiquidityPool.sol
hieronx: It is the owner of the request, that is the idea of the latest iteration of the 7540 spec. That when you submit a request, you specify a receiver, which "receives the request" and is then the owner of the request going forward. I think this makes sense. Also because the param names of these max* methods in the 4626 spec are owner.

hieronx: I agree we can possibly change this in the investment manager implementation, but in liquidity pool I don't think we should deviate from the 7540 & 4626 specs.

ilinzweilin (dismissed):

src/InvestmentManager.sol
hieronx: Yes I was just making the ordering consistent with `InvestmentManager.requestRedeem` and the ordering of params in `LiquidityPool`.

NunoAlexandre (approved):

PR #217, opened 2023-11-29
Transfer proxy factory

hieronx:


src/factories/TransferProxyFactory.sol
```diff
@@ -0,0 +1,59 @@
...
+    address public immutable recoverer;
+
+    constructor(address poolManager_, bytes32 destination_, address recoverer_) {
+        poolManager = IPoolManager(poolManager_);
+        destination = destination_;
+        recoverer = recoverer_;
+    }
+
+    /// @inheritdoc ITransferProxy
+    function transfer(address asset, uint128 amount) external {
```
AStox, line 20: What do you think about adding a `destination` argument here as well, and revert if it's not equal the set destination on the contract as a sanity check. Otherwise I dont think many people will actually verify the destination before transferring.

AStox (approved):
Looks good! Just left an idea

src/factories/TransferProxyFactory.sol
hieronx: I don't think that's worth the gas cost, should just be abstracted away in the UI

ilinzweilin (approved):

PR #220, opened 2023-12-14
change deployer permissions

ilinzweilin:


script/Deployer.sol
```diff
@@ -50,13 +53,17 @@ contract Deployer is Script {
...
+        AuthLike(restrictionManagerFactory).rely(address(poolManager));
+
+        AuthLike(liquidityPoolFactory).rely(address(root));
+        AuthLike(trancheTokenFactory).rely(address(root));
+        AuthLike(restrictionManagerFactory).rely(address(root));
 
-        LiquidityPoolFactory(liquidityPoolFactory).rely(address(root));
-        TrancheTokenFactory(trancheTokenFactory).rely(address(root));
-        RestrictionManagerFactory(restrictionManagerFactory).rely(address(root));
+        AuthLike(liquidityPoolFactory).deny(deployer);
```
hieronx, line 64: Why not put these in `removeDeployerAccess`?

script/Deployer.sol
```diff
@@ -92,11 +99,12 @@ contract Deployer is Script {
...
-        RouterLike(router).deny(deployer);
+        AuthLike(router).deny(deployer);
         root.deny(deployer);
         investmentManager.deny(deployer);
         poolManager.deny(deployer);
         escrow.deny(deployer);
+        userEscrow.deny(deployer);
         gateway.deny(deployer);
         pauseAdmin.deny(deployer);
         delayedAdmin.deny(deployer);
```
hieronx, line 110: Looks like router is also missing here?

ilinzweilin: the router is the first one that is denied

script/Deployer.sol
ilinzweilin: the factories are not global vars

hieronx: Let's make them global vars then. There's two reasons I think it should be in `removeDeployerAccess`:
1. This method is not called in the permissionless deploy script. This is quite helpful for testing, we can deploy while keeping the deployer access.
2. It is easier to audit that all `deny`s are done.

script/Deployer.sol
hieronx: Ah true!

hieronx (approved):

PR #221, opened 2023-12-14
Fix admin setup & ERC-7575 interface ID

hieronx:


AStox (commented):
Looks good, my only critique would be to also update `Permissionless.s.sol` with the changes added to `Axelar.s.sol`. Probably we'll never use it again, but I don't think we should have out of date code in our repo like that. Either that or we should just delete it altogether.

ilinzweilin (approved):

PR #232, opened 2024-01-03
Simplify Axelar router

hieronx:


src/gateway/routers/axelar/Router.sol
```diff
@@ -73,7 +59,10 @@ contract AxelarRouter is Auth {
         string calldata sourceChain,
         string calldata sourceAddress,
         bytes calldata payload
-    ) public onlyCentrifugeChainOrigin(sourceChain, sourceAddress) {
```
hieronx, line 76: Removing modifier as there's little value in this if it's not reused, and it makes the code easier to read.

src/gateway/routers/axelar/Router.sol
```diff
@@ -23,7 +23,9 @@ interface GatewayLike {
 /// @notice Routing contract that integrates with an Axelar Gateway
 contract AxelarRouter is Auth {
     string public constant CENTRIFUGE_CHAIN_ID = "centrifuge";
-    string public constant CENTRIFUGE_CHAIN_ADDRESS = "0x7369626CEF070000000000000000000000000000";
+    bytes32 public constant CENTRIFUGE_CHAIN_ID_HASH = keccak256(bytes("centrifuge"));
```
hieronx, line 26: Hashing already so we don't need to do that on every `execute` call

PR #237, opened 2024-01-26
Remove decrease logic, add cancelation view methods

hieronx:


AStox:
Looks good to me so far, but whats the reasoning behind this change? Users that want to decrease orders will have to cancel first, and then re-order?

AStox (approved):

AStox (approved):

src/InvestmentManager.sol
```diff
@@ -352,6 +343,8 @@ contract InvestmentManager is Auth {
         state.maxWithdraw = state.maxWithdraw + currencyPayout;
         state.pendingDepositRequest = remainingInvestOrder;
 
+        if (remainingInvestOrder == 0) state.pendingCancelDepositRequest = false;
```
ilinzweilin, line 346: I think we have to set ` state.pendingCancelDepositRequest = false;`  also in the `handleExecutedCollectInvest` function. Maybe the order got already fulfilled and `remainingInvestOrder=0` and there is nothing to cancel anymore... same for `handleExecutedCollectInvest`

ilinzweilin (commented):
In general the logic of the PR looks good, but I don't think we should rely on the pendingRequest variables and block investros' actions based on that. I think cfg chain has to rather handle multiple cancellation requests correctly in case an investor would fire them multiple times. Otherwise investors might get blocked from requesting deposits / redemptions, in case messages cancellation messages get lost / delayed. There is no downside on the evm part to cancel an order multiple times, or even request a deposit while your prev order is being cancelled... In fact, investors might want to cancel a prev big order, for example and in the meantime lock in a smaller order.

PR #239, opened 2024-02-19
Use bitmap to optimize boolean storage

hieronx:


AStox (commented):
This is a cool one! Is the reason for this purely gas savings or are there any other benefits?

(Also, looks like tests are failing)

PR #243, opened 2024-03-01
Message parsing refactor (part 2)

hieronx:


src/gateway/routers/axelar/Router.sol
```diff
@@ -51,9 +63,20 @@ contract AxelarRouter {
...
+        axelarGasService.payNativeGasForContractCall{value: msg.value}(
+            sender, CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, payload, sender
+        );
+    }
+
+    function send(bytes calldata paypload) public {
         require(msg.sender == address(aggregator), "AxelarRouter/only-aggregator-allowed-to-call");
 
-        axelarGateway.callContract(CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, message);
+        axelarGateway.callContract(CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, paypload);
```
AStox, line 80:
```suggestion
        axelarGateway.callContract(CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, payload);
```

AStox (commented):
For the auto gas payments I'm a bit out of my comfort zone so I'll read about how gas payments are made on axelar and re-review.

For message parsing refactor, it all looks good but I'm wondering what the goal is with this refactor. While I like the idea of each manager contract handling their own messages , it seems to me like we've just shifted the complexity of message parsing from its own self-contained library into the manager contracts, bloating those contracts. And although it's more gas efficient, I think Gateway's `handle` is actually less readable now.

hieronx:
Closing in favour of https://github.com/centrifuge/liquidity-pools/pull/261, that removes the gas logic

PR #262, opened 2024-03-13
Recover Tokens

AStox:
This PR adds functionality for governance to recover tokens that were accidentally sent to the wrong contract


src/libraries/MessagesLib.sol
```diff
@@ -504,6 +506,18 @@ library MessagesLib {
         return (_msg.toBytes32(1));
     }
 
+    /**
+    * Recover Tokens sent to the wrong contract
+    *
+    * 0: call type (uint8 = 1 byte)
+    * 1-20: The contract address (address = 20 bytes)
+    * 21-40: The recipient address (address = 20 bytes)
```
hieronx, line 514: These should be 32 bytes, of which the first 20 bytes are parsed as an address, for compatibility with future non EVM use cases

src/LiquidityPool.sol
```diff
@@ -90,6 +90,10 @@ contract LiquidityPool is Auth, IERC7540 {
         emit File(what, data);
     }
 
+    function recoverTokens(address to, uint256 amount) external auth {
```
hieronx, line 93: We should add asset as another param. Both for non liquidity pool contracts we add it to, but even for liquidity pool, someone can accidentally send USDC to the FRAX liquidity pool

hieronx (commented):
Nice work!

src/LiquidityPool.sol
AStox: Great point. I'll update it

src/LiquidityPool.sol
```diff
@@ -90,6 +90,10 @@ contract LiquidityPool is Auth, IERC7540 {
         emit File(what, data);
     }
 
+    function recoverTokens(address to, address token, uint256 amount) external auth {
```
hieronx, line 93: Would be cleaner if the order of the params of the method matches the order of the transfer func
```suggestion
    function recoverTokens(address token, address to, uint256 amount) external auth {
```

src/Root.sol
```diff
@@ -30,6 +34,7 @@ contract Root is Auth {
     event CancelRely(address indexed target);
     event RelyContract(address indexed target, address indexed user);
     event DenyContract(address indexed target, address indexed user);
+    event RecoverTokens(address indexed fromContract, address indexed toUser, address indexed token, uint256 amount);
```
hieronx, line 37: Compared to other messages, plus using the transfer method ordering, I would name it like this now:
```suggestion
    event RecoverTokens(address indexed target, address indexed token, address indexed to, uint256 amount);
```

test/integration/Admin.t.sol
```diff
@@ -260,4 +260,22 @@ contract AdminTest is BaseTest {
         root.relyContract(address(investmentManager), address(this));
         assertEq(investmentManager.wards(address(this)), 1);
     }
+
+    //------ Token Recovery tests ------///
+    function testRecoverTokensFromLiquidityPool() public {
+        deploySimplePool();
+        address clumsyUser = vm.addr(0x1234);
```
hieronx, line 267: 😄

src/gateway/Gateway.sol
```diff
@@ -90,6 +90,7 @@ interface RootLike {
     function paused() external returns (bool);
     function scheduleRely(address target) external;
     function cancelRely(address target) external;
+    function recoverTokens(address fromContract, address toUser, address token, uint256 amount) external;
```
hieronx, line 93:
```suggestion
    function recoverTokens(address target, address token, address to, uint256 amount) external;
```

test/mocks/MockCentrifugeChain.sol
```diff
@@ -159,6 +159,17 @@ contract MockCentrifugeChain is Test {
         _execute(_message);
     }
 
+    function recoverTokens(address fromContract, address toUser, address token, uint256 amount) public {
+        bytes memory _message = abi.encodePacked(
+            uint8(MessagesLib.Call.RecoverTokens),
+            bytes32(bytes20(fromContract)),
+            bytes32(bytes20(toUser)),
+            bytes32(bytes20(token)),
```
hieronx, line 167: There's a helper for this in `BytesLib`:
```suggestion
            fromContract.toBytes32(),
            toUser.toBytes32(),
            token.toBytes32(),
```

test/mocks/MockCentrifugeChain.sol
```diff
@@ -159,6 +159,17 @@ contract MockCentrifugeChain is Test {
         _execute(_message);
     }
 
+    function recoverTokens(address fromContract, address toUser, address token, uint256 amount) public {
```
hieronx, line 162: You haven't swapped the param names here, but are actually calling this with token before toUser  already from the test
```suggestion
    function recoverTokens(address target, address token, address to, uint256 amount) public {
```

hieronx (approved):
Looks good!

hieronx (dismissed):

src/gateway/Gateway.sol
```diff
@@ -93,6 +94,8 @@ contract Gateway is Auth {
             manager = poolManager;
         } else if (id == 27) {
             manager = investmentManager;
+        } else if (id == 31) {
```
ilinzweilin, line 97: I would add it as an "or" statement and combine with the other root cases

ilinzweilin (approved):

ilinzweilin (approved):

PR #264, opened 2024-03-20
Add vault lookup, separate cancelation flow

hieronx:


src/PoolManager.sol
```diff
@@ -527,11 +523,11 @@ contract PoolManager is Auth {
     }
 
     function getLiquidityPool(uint64 poolId, bytes16 trancheId, uint128 currencyId) public view returns (address) {
-        return pools[poolId].tranches[trancheId].liquidityPools[currencyIdToAddress[currencyId]];
+        return TrancheTokenLike(pools[poolId].tranches[trancheId].token).vault(currencyIdToAddress[currencyId]);
```
AStox, line 526: Bit more convoluted than before, but it seems like a necessary change to conform to ERC7575?

src/token/Tranche.sol
```diff
@@ -32,8 +33,13 @@ contract TrancheToken is ERC20 {
 
     mapping(address => bool) public trustedForwarders;
 
+    /// @dev Look up vault by the asset (part of ERC7575)
+    mapping(address asset => address) public vault;
```
AStox, line 37: I'm not a huge fan of referring to liquidity pools by another name here (although I understand its a requirement for the ERC). Should we consider renaming liquidity pools to vaults?

AStox (commented):
I wasnt part of the conversation behind this one, so I'm trying to work backwards to understand this PR. Seems like we're:
1) consolidating the erc7575 interface
2) restructuring how we keep track of liquidity pools in order to conform with erc7575
Is that accurate?

src/PoolManager.sol
hieronx: Yes exactly, this is part of ERC7575. I do think this is beneficial for integrations, they can easily find a vault (liquidity pool) for any asset, from the tranche token contract.

AStox (dismissed):

ilinzweilin (approved):

PR #266, opened 2024-03-24
Use deltas for reducing pending requests

hieronx:


src/InvestmentManager.sol
```diff
@@ -358,7 +361,10 @@ contract InvestmentManager is Auth {
             ((maxRedeem(liquidityPool, user)) + trancheTokenPayout).toUint128()
         );
         state.maxWithdraw = state.maxWithdraw + currencyPayout;
-        state.pendingRedeemRequest = remainingRedeemOrder;
+        state.pendingRedeemRequest =
+            state.pendingRedeemRequest > fulfilledRedeemOrder ? state.pendingRedeemRequest - fulfilledRedeemOrder : 0;
```
ilinzweilin, line 365: wouldn't you simply use the trancheTokenPayout for that? 
Why do we need the `fulfilledRedeemOrder` variable

src/InvestmentManager.sol
```diff
@@ -328,7 +328,10 @@ contract InvestmentManager is Auth {
             liquidityPool, _maxDeposit(liquidityPool, user) + currencyPayout, state.maxMint + trancheTokenPayout
         );
         state.maxMint = state.maxMint + trancheTokenPayout;
-        state.pendingDepositRequest = remainingInvestOrder;
+        state.pendingDepositRequest =
```
ilinzweilin, line 331: wouldn't you simply use the currencyPayout for that? 
Why do we need the `fulfilledInvestOrder` variable?

ilinzweilin (changes requested):

AStox (commented):
Looks good to me!

PR #275, opened 2024-04-05
Add price per share, clean up price calculation

hieronx:


src/ERC7540Vault.sol
```diff
@@ -293,8 +295,13 @@ contract ERC7540Vault is Auth, IERC7540 {
     }
 
     // --- Helpers ---
-    function exchangeRateLastUpdated() external view returns (uint64) {
-        return manager.exchangeRateLastUpdated(address(this));
+    /// @notice Price of 1 unit of share, quoted in the decimals of the asset
+    function pricePerShare() external view returns (uint256) {
```
hieronx, line 299: This methods enables looking up the price, without also needing to look up the decimals of the tranche tokens (thus also saving gas).

src/InvestmentManager.sol
```diff
@@ -586,20 +586,14 @@ contract InvestmentManager is Auth, IInvestmentManager {
     }
 
     function _calculatePrice(address vault, uint128 assets, uint128 shares) internal view returns (uint256 price) {
```
hieronx, line 588: We don't need the separate anymore since we refactored the decrease/cancelation logic in a previous PR

src/ERC7540Vault.sol
```diff
@@ -34,6 +34,7 @@ contract ERC7540Vault is Auth, IERC7540 {
     /// @notice The restricted ERC-20 vault share (tranche token).
     ///         Has a ratio (token price) of underlying assets exchanged on deposit/mint/withdraw/redeem.
     address public immutable share;
+    uint8 public immutable shareDecimals;
```
hieronx, line 37: I'm wondering whether we should also store `assetDecimals`, and then always pass `assetDecimals` + `shareDecimals` on every call to `InvestmentManager`. This would save a lot of storage lookups since almost all methods in `InvestmentManager` need to know these decimals.

TBD, let me know what you think!

AStox: I cant think of any reason not to!

AStox (approved):
👍

ilinzweilin (approved):

PR #277, opened 2024-04-16
Add `CentrifugeRouter`

ilinzweilin:


src/LiquidityPoolRouter.sol
```diff
@@ -0,0 +1,74 @@
...
+    function redeem(uint256 shares, address receiver, address owner) external returns (uint256 assets);
+    function asset() external returns (address asset);
+    function share() external returns (address asset);
+}
+
+interface ERC20Like {
+    function transferFrom(address from, address to, uint256 amount) external returns (bool);
+}
+
+contract LiquidityPoolRouter is Auth {
```
hieronx, line 22:
```suggestion
contract CentrifugeRouter is Auth {
```

src/LiquidityPoolRouter.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.21;
+
+import {Auth} from "./Auth.sol";
+import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
+
+interface LiquidityPoolLike {
```
hieronx, line 7: You can import `IERC7540` instead

src/LiquidityPoolRouter.sol
```diff
@@ -0,0 +1,74 @@
...
+    // --- Interactions ---
+
+    // --- Deposit ---
+    function requestDeposit(address liquidityPool, uint256 amount) external {
+        ERC20Like(LiquidityPoolLike(liquidityPool).asset()).transferFrom(msg.sender, address(this), amount);
+        LiquidityPoolLike(liquidityPool).requestDeposit(amount, msg.sender, address(this), "");
+    }
+
+    function lockDepositRequest(address liquidityPool, uint256 amount) external {
+        ERC20Like(LiquidityPoolLike(liquidityPool).asset()).transferFrom(msg.sender, address(this), amount);
```
hieronx, line 34: This should use `SafeTransferLib.safeTransferFrom`

src/LiquidityPoolRouter.sol
```diff
@@ -0,0 +1,74 @@
...
+}
+
+contract LiquidityPoolRouter is Auth {
+    mapping(address => mapping(address => uint256)) public lockedRequests;
+
+    // --- Interactions ---
+
+    // --- Deposit ---
+    function requestDeposit(address liquidityPool, uint256 amount) external {
+        ERC20Like(LiquidityPoolLike(liquidityPool).asset()).transferFrom(msg.sender, address(this), amount);
```
hieronx, line 29: This should use `SafeTransferLib.safeTransferFrom`

src/LiquidityPool.sol
```diff
@@ -258,9 +279,10 @@ contract LiquidityPool is Auth, IERC7540 {
     }
 
     /// @inheritdoc IERC7575
-    function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
-        shares = manager.deposit(address(this), assets, receiver, msg.sender);
-        emit Deposit(msg.sender, receiver, assets, shares);
+    function deposit(uint256 assets, address receiver, address owner) external returns (uint256 shares) {
```
hieronx, line 282: We need to also keep the `deposit(assets, receiver)` method, and add an overload for the one with `owner`

src/LiquidityPoolRouter.sol
```diff
@@ -0,0 +1,74 @@
...
+        LiquidityPoolLike(liquidityPool).requestDeposit(amount, msg.sender, address(this), "");
```
hieronx, line 30: I think this should require an approval?

src/LiquidityPoolRouter.sol
```diff
@@ -0,0 +1,74 @@
...
+    function lockDepositRequest(address liquidityPool, uint256 amount) external {
+        ERC20Like(LiquidityPoolLike(liquidityPool).asset()).transferFrom(msg.sender, address(this), amount);
+        lockedRequests[msg.sender][liquidityPool] += amount;
+    }
+
+    function unlockDepositRequest(address liquidityPool) external {
+        uint256 lockedRequest = lockedRequests[msg.sender][liquidityPool];
+        require(lockedRequest > 0, "LiquidityPoolRouter/investor-has-no-locked-balance");
+        lockedRequests[msg.sender][liquidityPool] = 0;
+        ERC20Like(LiquidityPoolLike(liquidityPool).asset()).transferFrom(msg.sender, address(this), lockedRequest);
```
hieronx, line 42: I think this should use `SafeTransferLib.transfer` (rather than transferFrom, since it otherwise wouldn't work with USDC)

src/LiquidityPoolRouter.sol
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    function executeLockedRequestDeposit(address liquidityPool, address investor) external {
```
hieronx, line 45:
```suggestion
    function executeLockedDepositRequest(address liquidityPool, address investor) external {
```

src/LiquidityPool.sol
```diff
@@ -94,6 +99,22 @@ contract LiquidityPool is Auth, IERC7540 {
...
+        emit Endorse(user);
+    }
+
+    /// @dev remove endorsement
+    function veto(address user) external auth {
+        endorsements[user] = 0;
+        emit Veto(user);
+    }
+
+    function isEndorsed(address user) public view returns (bool) {
```
hieronx, line 114: More consistent with `auth`
```suggestion
    function endorsed(address user) public view returns (bool) {
```

src/LiquidityPool.sol
```diff
@@ -295,7 +317,7 @@ contract LiquidityPool is Auth, IERC7540 {
     /// @inheritdoc IERC7575
     /// @notice     DOES NOT support owner != msg.sender since shares are already transferred on requestRedeem
```
hieronx, line 318:
```suggestion
```

hieronx (commented):
I realize it's still WIP but just leaving a few notes as I read through it. Overall looks great!!

src/ERC7540Vault.sol
```diff
@@ -234,9 +255,19 @@ contract ERC7540Vault is Auth, IERC7540 {
...
 
     /// @inheritdoc IERC7575
+
     function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
-        shares = manager.deposit(address(this), assets, receiver, msg.sender);
-        emit Deposit(msg.sender, receiver, assets, shares);
+        shares = _deposit(assets, receiver, msg.sender);
+    }
+
+    function deposit(uint256 assets, address receiver, address owner) external returns (uint256 shares) {
```
peculiarity, line 263: Not quite sure why we have 2 methods. Can we leave only this one? Most of the methods rely on passing the owner. Also If for some reason there is a need for both methods, probably there should be one for redeem as well? Which these changes there is only one `redeem` method that accepts `owner`.

test/integration/Deposit.t.sol
```diff
@@ -94,9 +94,14 @@ contract DepositTest is BaseTest {
...
 
+        vm.assume(randomUser != self);
         // deposit 50% of the amount
-        vault.deposit(amount / 2, self); // mint half the amount
+        vm.startPrank(randomUser); // try to claim deposit on behalf of user and set the wrong user as receiver
+        vm.expectRevert(bytes("LiquidityPool/not-owner-or-endorsed"));
+        vault.deposit(amount / 2, randomUser, self);
+        vm.stopPrank();
 
+        vault.deposit(amount / 2, self, self); // deposit half the amount
```
peculiarity, line 104: I will put the comment here but it applies to all occurrences of `vault.deposit`. You replaced "old" signature with the "new" one everywhere. Doing so, there is no test for the `deposit(uint256 assets, address receiver)` . I can break it and there is nothing that will catch it.
I literally hard-coded the return value and nothing broke for me:
```
  function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
        return 9;
   }
```

Of course this won't be needed, if you decide to get rid of the "old" method.

peculiarity (commented):
There are also 2 failing tests.
1. `PoolManager.t.sol:testRestrictionSetWithIntegration` - I'm not aware of the details of the suite but since you added a new field there, probably something must be added in the setup.
2. `Redeem.t.sol:testRedeeemWithApproval` - I think it's the change  where you included the `enderesed(msg.sender)` because it reverts with ` revert: LiquidityPool/not-owner-or-endorsed`

src/CentrifugeRouter.sol
```diff
@@ -0,0 +1,70 @@
...
+    mapping(address => mapping(address => uint256)) public lockedRequests;
+
+    // --- Interactions ---
+
+    // --- Deposit ---
+    function requestDeposit(address vault, uint256 amount) external {
+        SafeTransferLib.safeTransferFrom(IERC7540(vault).asset(), msg.sender, address(this), amount);
+        IERC20Like(IERC7540(vault).asset()).approve(vault, amount); // option to add approve function, make poolManager
+            // ward and create global approval on vault deployment
+        IERC7540(vault).requestDeposit(amount, msg.sender, address(this), "");
```
hieronx, line 26: This should work, because the router is endorsed (aka automatically an operator):
```suggestion
        IERC7540(vault).requestDeposit(amount, msg.sender, msg.sender, "");
```

src/CentrifugeRouter.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.21;
+
+import {Auth} from "./Auth.sol";
+import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
+import "./interfaces/IERC7540.sol";
+
+interface IERC7540ExtLike {
```
hieronx, line 8: The interface was updated already, you can just use `IERC7540`

src/ERC7540Vault.sol
```diff
@@ -45,8 +49,14 @@ contract ERC7540Vault is Auth, IERC7540 {
     /// @dev    Requests for Centrifuge pool are non-transferable and all have ID = 0
     uint256 constant REQUEST_ID = 0;
 
+    /// @notice Routers that are allowed to claim withdrawels / redemptions on behalf of investors
+    mapping(address => uint256) public endorsements;
```
hieronx, line 53: Maybe in a future PR, but we should move this to the investment manager

src/PoolManager.sol
```diff
@@ -51,6 +51,7 @@ contract PoolManager is Auth, IPoolManager {
     TrancheTokenFactoryLike public trancheTokenFactory;
     ERC7540VaultFactory public vaultFactory;
     RestrictionManagerFactoryLike public restrictionManagerFactory;
+    address public cfgRouter;
```
hieronx, line 54: nit: I'm not a fan of using the cfg abbreviation since it implies the CFG token. I'd just call this router

```suggestion
    address public router;
```

script/Deployer.sol
```diff
@@ -36,6 +37,7 @@ contract Deployer is Script {
     Guardian public guardian;
     Gateway public gateway;
     Aggregator public aggregator;
+    CentrifugeRouter public cfgRouter;
```
hieronx, line 40: nit: like the other, would call this router
```suggestion
    CentrifugeRouter public router;
```

src/PoolManager.sol
```diff
@@ -399,7 +401,11 @@ contract PoolManager is Auth, IPoolManager {
         address restrictionManager = restrictionManagerFactory.newRestrictionManager(
             undeployedTranche.restrictionSet, token, restrictionManagerWards
         );
+
         TrancheTokenLike(token).file("restrictionManager", restrictionManager);
+        if (cfgRouter != address(0)) {
```
hieronx, line 406: With the change I proposed in the router contract, the router just acts as an operator of the investor, and does not even need to be a member

hieronx: So if we do that + move endorsements to the investment manager, I think the pool manager doesn't even need to know about the router anymore

hieronx: Actually that's not true. It won't need to be a member anymore for "normal" requests, but it will need to be a member for locked requests, since then it needs to take custody of the assets

src/CentrifugeRouter.sol
```diff
@@ -0,0 +1,70 @@
...
+        IERC7540(vault).requestRedeem(amount, msg.sender, address(this), "");
+    }
+
+    function claimRedeem(address vault, address investor) external {
+        uint256 maxRedeem = IERC7540(vault).maxRedeem(investor);
+        require(maxRedeem > 0, "CentrifugeRouter/investor-has-no-balance-to-claim");
+        IERC7540(vault).redeem(maxRedeem, investor, investor);
+    }
+
+    // --- ERC20 Recovery ---
```
hieronx, line 66: Nit: in all the other contracts, we put this in the Administration section at the top (below the contructor). I would either do that, or move all the other methods to the bottom like here, for consistency.

src/InvestmentManager.sol
```diff
@@ -472,6 +475,11 @@ contract InvestmentManager is Auth, IInvestmentManager {
         (, lastUpdated) = poolManager.getTrancheTokenPrice(vault_.poolId(), vault_.trancheId(), vault_.asset());
     }
 
+    /// @inheritdoc IInvestmentManager
+    function isGlobalOperator(address, /* vault */ address user) public view returns (bool) {
```
hieronx, line 479: Doing it this way has 2 benefits
- Enables us to possibly swap the investment mgr in the future and add other cases for when a contract is an operator of another user, without having to swap the vaults
- Not adding a dependency of the vault on the root

test/integration/Deposit.t.sol
hieronx: ERC7540 specifies (intentionally, to support the old interface for claiming) that both overloaded deposit methods should be included. So we shouldn't remove one, but I agree we should definitely add tests for both versions!

src/interfaces/ICentrifugeRouter.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.5.0;
+
+interface ICentrifugeRouter {
```
peculiarity, line 4: I'm just curious what do we need this interface for? Is it so that it's easier for 3rd parties to integrate?  They just import the interface  in their code  and does something like:
```
ICentrifugeRouter(address).requestDepost(...)
```

test/integration/Deposit.t.sol
```diff
@@ -331,6 +336,67 @@ contract DepositTest is BaseTest {
...
+        address router = makeAddr("router");
+
+        vm.startPrank(router);
+        vm.expectRevert(bytes("ERC7540Vault/invalid-owner")); // fail without endorsement
+        vault.deposit(amount, receiver, address(this));
+        vm.stopPrank();
+
+        // endorse router
+        root.endorse(router);
+        vm.startPrank(router); // try to claim deposit on behalt of user and set the wrong user as receiver
```
peculiarity, line 390: There is a type `s/behalt/behalf`

src/interfaces/ICentrifugeRouter.sol
hieronx: Reason number 1 is that we have interfaces for all contracts such that we can have natspec documentation in the interfaces, and then use `inheritdoc` tags in the contracts.

Second reason is exactly what you said!

peculiarity (approved):

src/InvestmentManager.sol
AStox, line 479:
```suggestion
    function isGlobalOperator(address /* vault */, address user) public view returns (bool) {
```

AStox (approved):

PR #280, opened 2024-04-18
PoolManager and InvestmentManager interface setup

hieronx:


src/interfaces/IPoolManager.sol
```diff
@@ -0,0 +1,182 @@
...
+    );
+    event TransferTrancheTokensToEVM(
+        uint64 indexed poolId,
+        bytes16 indexed trancheId,
+        uint64 indexed destinationChainId,
+        address destinationAddress,
+        uint128 amount
+    );
+
+    /// @notice TODO
```
hieronx, line 67: We should fix these TODOs in a future PR

src/PoolManager.sol
```diff
@@ -12,6 +12,7 @@ import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
 import {MathLib} from "./libraries/MathLib.sol";
 import {MessagesLib} from "src/libraries/MessagesLib.sol";
 import {CastLib} from "src/libraries/CastLib.sol";
+import {Pool, Tranche, TrancheTokenPrice, UndeployedTranche, IPoolManager} from "src/interfaces/IPoolManager.sol";
 import {BytesLib} from "src/libraries/BytesLib.sol";
 
 interface GatewayLike {
```
hieronx, line 18: Once we add interfaces for other contracts, we can replace these inline interfaces with imports

AStox (approved):
Nice. Loving all these cleanup PRs. It would be nice to fill in all the TODO comments but obv not blocking.

ilinzweilin (approved):

PR #281, opened 2024-04-19
Refactor Aggregator contract

hieronx:


src/libraries/ArrayLib.sol
```diff
@@ -9,15 +9,17 @@ library ArrayLib {
         }
     }
 
-    function decreaseFirstNValues(uint16[8] storage arr, uint8 numValues, uint16 decrease) internal {
```
hieronx, line 12: This parameter was unused and only complicated the method needlessly

src/libraries/ArrayLib.sol
```diff
@@ -4,24 +4,26 @@ pragma solidity 0.8.21;
...
             if (numValues == 0) return;
 
             if (arr[i] > 0) {
-                arr[i] -= decrease;
+                arr[i] -= 1;
                 numValues--;
             }
         }
+
+        require(numValues == 0, "ArrayLib/invalid-values");
```
hieronx, line 22: This makes the method stricter, catching issues with trying to decrease by too much

script/Deployer.sol
```diff
@@ -77,6 +77,7 @@ contract Deployer is Script {
 
         // Wire aggregator
         aggregator.file("routers", routers);
+        aggregator.rely(address(gateway));
```
hieronx, line 80: Required now because it's changed from an `onlyGateway` modifier to an `auth` modifier

src/gateway/routers/Aggregator.sol
```diff
@@ -73,11 +72,12 @@ contract Aggregator is Auth, IAggregator {
...
         Router memory router = validRouters[msg.sender];
         require(router.id != 0, "Aggregator/invalid-router");
-        _handle(payload, router);
+        _handle(payload, router, false);
     }
 
-    function _handle(bytes calldata payload, Router memory router) internal {
+    function _handle(bytes calldata payload, Router memory router, bool isRecovery) internal {
         if (MessagesLib.isRecoveryMessage(payload)) {
+            require(!isRecovery, "Aggregator/no-recursive-recovery-allowed");
```
hieronx, line 80: To prevent doing a recovery of a recovery message

src/gateway/routers/Aggregator.sol
```diff
@@ -91,42 +91,37 @@ contract Aggregator is Auth, IAggregator {
...
-                    delete pendingMessages[messageHash];
-                }
+                gateway.handle(state.pendingMessage);
             } else {
                 gateway.handle(payload);
             }
 
+            // Only if there are no more pending messages, remove the pending message
+            if (state.votes.isEmpty()) {
+                delete state.pendingMessage;
```
AStox, line 119: I don't think there's any risk to just leaving the pendingMessage is there? We could save a little gas by skipping this.

AStox (approved):
Looks good. Much more elegant imo. Left one potential optimization

src/gateway/routers/Aggregator.sol
hieronx: You get gas refunds if you clear storage, so this actually is a gas optimization

AStox: Oh wow! Didn't know that

AStox (dismissed):

src/gateway/routers/Aggregator.sol
```diff
@@ -46,18 +47,28 @@ contract Aggregator is Auth, IAggregator {
...
         if (what == "routers") {
-            require(routers_.length <= MAX_ROUTER_COUNT, "Aggregator/exceeds-max-router-count");
+            uint8 quorum_ = uint8(routers_.length);
+            require(quorum_ > 0, "Aggregator/empty-router-set");
+            require(quorum_ <= MAX_ROUTER_COUNT, "Aggregator/exceeds-max-router-count");
+
+            uint64 sessionId = 0;
+            if (routers.length > 0) {
+                // Increment session id if it is not the initial router setup and the quorum was decreased
+                Router memory prevRouter = validRouters[address(routers[0])];
```
peculiarity, line 57: `routers[0]` returns an address so I don't think you have to explicitly cast it to an address :S

AStox (approved):

src/gateway/routers/Aggregator.sol
```diff
@@ -31,9 +33,8 @@ contract Aggregator is Auth, IAggregator {
 
     address[] public routers;
     mapping(address router => Router) public validRouters;
```
peculiarity, line 35: I know this is not the topic of this PR but I will just write down this suggestion. 
Can't remember where exactly I read about this ( code clean book or Martin Fowler or some where else ) but using `valid` and `invalid` is ambiguous. `Valid` doesn't show how this is valid. 
In our case probably a better name could be `approvedRouters` or `activeRouters` or `whitelistedRouters` .
This is just a side comment while I'm revising the code.

src/gateway/routers/Aggregator.sol
```diff
@@ -153,26 +173,28 @@ contract Aggregator is Auth, IAggregator {
...
-    function send(bytes calldata message) public {
-        require(msg.sender == address(gateway), "Aggregator/only-gateway-allowed-to-call");
-
+    function send(bytes calldata message) external auth {
         uint256 numRouters = routers.length;
         require(numRouters > 0, "Aggregator/not-initialized");
 
         bytes memory proof = abi.encodePacked(uint8(MessagesLib.Call.MessageProof), keccak256(message));
-        for (uint256 i; i < numRouters; ++i) {
+        for (uint256 i; i < numRouters; i++) {
```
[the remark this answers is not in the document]
hieronx: This changed with 0.8.22 (which we anyways should upgrade to): https://soliditylang.org/blog/2023/10/25/solidity-0.8.22-release-announcement/

src/gateway/routers/Aggregator.sol
hieronx: Yeah I agree, will change it to `activeRouters`, I like that

src/gateway/routers/Aggregator.sol
hieronx: True, good catch

src/interfaces/gateway/IAggregator.sol
```diff
@@ -2,20 +2,25 @@
...
-        // Max uint16 = 65,535 so at most 65,535 duplicate messages can be processed in parallel.
-        uint16[8] messages;
-        uint16[8] proofs;
+    struct Message {
+        /// @dev Counts are stored as integers (instead of boolean values) to accommodate duplicate
+        ///      messages (e.g. two investments from the same user with the same amount) being
+        ///      processed in parallel. The entire struct is packed in a single bytes32 slot.
+        ///      Max uint16 = 65,535 so at most 65,535 duplicate messages can be processed in parallel.
+        uint16[8] votes;
+        /// @notice Each time routers are updated, a new session starts which invalidates old votes
```
peculiarity, line 21: Actually that's not always true. A new sessions starts only if the new routers count ( quorum ) is less than the current one, isn't this correct?

src/gateway/routers/Aggregator.sol
```diff
@@ -70,80 +81,89 @@ contract Aggregator is Auth, IAggregator {
...
+        // Verify router and parse message hash
         bytes32 messageHash;
-        ConfirmationState storage state;
         if (isMessageProof) {
-            messageHash = MessagesLib.parseMessageProof(payload);
-            state = _confirmations[messageHash];
-            state.proofs[router.id - 1]++;
-
-            emit HandleProof(messageHash, msg.sender);
+            require(isRecovery || router.id != 1, "RouterAggregator/non-proof-router");
```
peculiarity, line 109:
```suggestion
            require(isRecovery || router.id !=PRIMARY_ROUTER_ID, "RouterAggregator/non-proof-router");
```

src/libraries/ArrayLib.sol
peculiarity:
```
 function dfnv(uint8[10] storage arr, uint8 numValues) internal {
        require(numValues <= arr.length, "Out-of-bound");
        if (numValues == 0) return;

        for (uint256 i; i < numValues; i++) {   
            if (arr[i] > 0) {
                arr[i] -= 1;
            }
        }
    }
    ```

ilinzweilin (dismissed):

PR #291, opened 2024-04-26
Batch message support

hieronx:
This PR introduces batched messaging. A new crosschain message type that contains multiple submessages to be executed sequentially.


src/gateway/Gateway.sol
```diff
@@ -77,6 +77,10 @@ contract Gateway is Auth, IGateway {
 
     // --- Incoming ---
     function handle(bytes calldata message) external auth pauseable {
+        _handle(message);
```
AStox, line 80: Whats the point of having `handle()` call `_handle()` with the same args?

AStox (commented):
It's being implicitly tested since it's in `BaseTest` now, but I think it would be nice to add a couple explicit unit tests to `Messages.t.sol` for batched messages

src/gateway/Gateway.sol
```diff
@@ -204,6 +204,23 @@ contract Gateway is Auth, IGateway {
             manager = investmentManager;
         } else if (id >= 21 && id <= 22 || id == 31) {
             manager = address(root);
+        } else if (id == 33) {
+            require(!isBatched, "Gateway/no-recursive-batching-allowed");
+            // Handle batch messages
```
hieronx, line 209: nit:
```suggestion
            // Handle batch messages
            require(!isBatched, "Gateway/no-recursive-batching-allowed");
```

peculiarity (approved):

PR #298, opened 2024-05-01
Add operator logic

hieronx:


src/interfaces/IERC20.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.21;
+pragma solidity >=0.5.0;
```
hieronx, line 2: Requested by Superform, this helps integrations use the interfaces without being locked to solidity 0.8.21

AStox (dismissed):
I would add a redemption integration test for good measure too, but otherwise it looks good! 👍

hieronx:
> I would add a redemption integration test

@AStox added, which actually caught a small issue :)

AStox (dismissed):
Haha I might've missed that issue, but glad my suggestion caught it!

AStox (approved):
Nice 👍

ilinzweilin (approved):

PR #304, opened 2024-05-16
Router updates

AStox:
This PR adds integration tests to the centrifuge router PR as well as a Multicall contract to allow the router functions to be multicalled.

Context: https://github.com/centrifuge/liquidity-pools/issues/247


src/CentrifugeRouter.sol
```diff
@@ -17,20 +19,23 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
 
     // --- Deposit ---
+
     /// @inheritdoc ICentrifugeRouter
-    function requestDeposit(address vault, uint256 amount) external {
-        IERC7540(vault).requestDeposit(amount, msg.sender, msg.sender);
+    function requestDeposit(address vault, uint256 amount) public {
+        SafeTransferLib.safeTransferFrom(IERC7540(vault).asset(), msg.sender, address(this), amount);
+        IERC20(IERC7540(vault).asset()).approve(vault, amount);
+        IERC7540(vault).requestDeposit(amount, msg.sender, address(this));
```
hieronx, line 27: Why is this reverted to the old logic?

src/CentrifugeRouter.sol
```diff
@@ -17,20 +19,23 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
     }
 
     /// @inheritdoc ICentrifugeRouter
-    function lockDepositRequest(address vault, uint256 amount) external {
+    function lockDepositRequest(address vault, uint256 amount) public {
```
hieronx, line 31: These can remain `external` now, right?

src/CentrifugeRouter.sol
```diff
@@ -43,7 +48,8 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         uint256 lockedRequest = lockedRequests[user][vault];
         require(lockedRequest > 0, "CentrifugeRouter/user-has-no-balance");
         lockedRequests[user][vault] = 0;
-        IERC7540(vault).requestDeposit(lockedRequest, msg.sender, address(this));
+        IERC20(IERC7540(vault).asset()).approve(vault, lockedRequest);
+        IERC7540(vault).requestDeposit(lockedRequest, user, address(this));
```
hieronx, line 52: Likewise why is this reverted?

src/Multicall.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.21;
+
+import "src/interfaces/IMulticall.sol";
+
+/// @title Multicall
+/// @notice Enables calling multiple methods in a single call to the contract
```
hieronx, line 7:
```suggestion
/// @notice Enables calling multiple methods in a single call to the contract
/// @author Uniswap v3 Periphery (contracts/base/Multicall.sol)
```

src/interfaces/IMulticall.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.21;
```
hieronx, line 2:
```suggestion
// SPDX-License-Identifier: AGPL-3.0-only
pragma solidity >=0.5.0;
```

src/CentrifugeRouter.sol
AStox: Maybe I missed this conversation somewhere, but without these lines the user has to explicitly approve the vault themselves, which I thought was something we were trying to avoid with this contract since its another UX focused contract.

src/CentrifugeRouter.sol
AStox: This one needs to stay like this in order to let anybody call `executeLockedDepositRequest`

hieronx: Or integrate the above into `PoolManager.deployVault`, but that is pretty tight coupling

src/CentrifugeRouter.sol
```diff
@@ -3,10 +3,12 @@ pragma solidity 0.8.21;
 
 import {Auth} from "src/Auth.sol";
 import {IERC7540} from "src/interfaces/IERC7540.sol";
+import {IERC20} from "src/interfaces/IERC20.sol";
 import {SafeTransferLib} from "src/libraries/SafeTransferLib.sol";
 import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
+import {Multicall} from "src/Multicall.sol";
 
-contract CentrifugeRouter is Auth, ICentrifugeRouter {
+contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
```
hieronx, line 11: This is unrelated to this PR, but something I noticed, I wonder if we need these lines:
```
require(maxDeposit > 0, "CentrifugeRouter/user-has-no-balance-to-claim");
```

in claimDeposit/Redeem. It is anyways checked in `deposit` itself already. It is minimal but would still be a small gas saving.

hieronx: Also this was included in the spec but missed in the implementation so far, maybe you want to add this as well
```solidity
  // --- View methods ---
  function getVault(uint64 poolId, bytes16 trancheId, address asset) external view returns (address) {
    // TODO
  }
```

This will require initialising it with the pool manager.

src/CentrifugeRouter.sol
AStox: Riiiight. Great idea! Multicall adds a lot of potential :)

src/CentrifugeRouter.sol
AStox: Updated!

src/CentrifugeRouter.sol
```diff
@@ -18,6 +18,11 @@ contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
         SafeTransferLib.safeTransfer(token, to, amount);
     }
 
+    // --- Approval ---
+    function approveVault(address vault) external {
```
hieronx, line 29: This should have an `inheritdoc` that links to `ICentrifugeRouter`

test/integration/CentrifugeRouter.t.sol
```diff
@@ -186,6 +188,34 @@ contract CentrifugeRoutertest is BaseTest {
...
+        ERC7540Vault vault = ERC7540Vault(vault_);
+        vm.label(vault_, "vault");
+
+        erc20.mint(self, amount);
+        erc20.approve(address(centrifugeRouter), amount);
+        centrifugeChain.updateMember(vault.poolId(), vault.trancheId(), self, type(uint64).max);
+        centrifugeRouter.lockDepositRequest(vault_, amount);
+
+        // multicall
+        bytes[] memory calls = new bytes[](2);
```
hieronx, line 205: Nice!

src/CentrifugeRouter.sol
```diff
@@ -2,20 +2,34 @@
...
+import {IERC7540} from "src/interfaces/IERC7540.sol";
+import {IERC20} from "src/interfaces/IERC20.sol";
 import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
+import {IPoolManager} from "src/interfaces/IPoolManager.sol";
 
-contract CentrifugeRouter is Auth, ICentrifugeRouter {
+contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
     /// @inheritdoc ICentrifugeRouter
     mapping(address user => mapping(address vault => uint256 amount)) public lockedRequests;
+    address public immutable poolManager;
```
hieronx, line 15: nit: we always put immutables on top
```suggestion
    address public immutable poolManager;
    
    /// @inheritdoc ICentrifugeRouter
    mapping(address user => mapping(address vault => uint256 amount)) public lockedRequests;
```

src/CentrifugeRouter.sol
```diff
@@ -2,20 +2,34 @@
...
+
+    constructor(address poolManager_) {
+        poolManager = poolManager_;
+    }
 
     // --- Administration ---
     /// @inheritdoc ICentrifugeRouter
```
hieronx, line 22: We should add a `file` method for the pool manager. Since the Router is actually somewhat hard to migrate with the locked requests state.

test/unit/CentrifugeRouter.t.sol
```diff
@@ -5,4 +5,12 @@ import "test/BaseTest.sol";
 import "src/interfaces/IERC7575.sol";
 import "src/interfaces/IERC7540.sol";
 
-contract CentrifugeRoutertest is BaseTest {}
+contract CentrifugeRoutertest is BaseTest {
```
hieronx, line 8:
```suggestion
contract CentrifugeRouterTest is BaseTest {
```

src/CentrifugeRouter.sol
```diff
@@ -2,20 +2,47 @@
...
+    }
+
     // --- Administration ---
     /// @inheritdoc ICentrifugeRouter
     function recoverTokens(address token, address to, uint256 amount) external auth {
         SafeTransferLib.safeTransfer(token, to, amount);
     }
 
+    /// @inheritdoc ICentrifugeRouter
+    function file(bytes32 what, address data) external auth {
```
hieronx, line 32: This is missing a `File` event

hieronx (commented):
I think this is my final comment ,looking really good!

src/CentrifugeRouter.sol
AStox: Updated!

hieronx: The event should be defined in the interface though :)

AStox: Fixed!

hieronx (approved):

src/CentrifugeRouter.sol
```diff
@@ -43,14 +68,13 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         uint256 lockedRequest = lockedRequests[user][vault];
         require(lockedRequest > 0, "CentrifugeRouter/user-has-no-balance");
         lockedRequests[user][vault] = 0;
-        IERC7540(vault).requestDeposit(lockedRequest, msg.sender, address(this));
+        IERC7540(vault).requestDeposit(lockedRequest, user, address(this));
```
peculiarity, line 71: nit: Probably we should keep the terminology as it is and use terms such as `owner` , `receiver` and in general variable names used across other contracts.

peculiarity (approved):

PR #305, opened 2024-05-18
Clean up ERC20, remove callbacks from ERC7540

hieronx:
Some improvements based on the findings in this audit report of the Maker ERC20 implementation: https://github.com/makerdao/nst/blob/dev/audit/cantina-report-review-makerdao-nst.pdf

As well as moving the `transferFrom` logic to an internal method as the `authTransferFrom` method is reusing this.

Finally, removes callbacks from the ERC7540 vault methods.


src/InvestmentManager.sol
```diff
@@ -26,10 +26,10 @@ interface VaultLike is IERC20 {
     function trancheId() external view returns (bytes16);
     function asset() external view returns (address);
     function share() external view returns (address);
-    function emitDepositClaimable(address owner, uint256 assets, uint256 shares) external;
-    function emitRedeemClaimable(address owner, uint256 assets, uint256 shares) external;
-    function emitCancelDepositClaimable(address owner, uint256 assets) external;
-    function emitCancelRedeemClaimable(address owner, uint256 shares) external;
+    function onDepositClaimable(address owner, uint256 assets, uint256 shares) external;
```
hieronx, line 29: A bit nicer imo since in theory this is a generic hook, that the vault can decide to implement and do whatever they want with, which now is just emitting an event

AStox (approved):
Looks good!

ilinzweilin (approved):

PR #306, opened 2024-05-21
refactor: Improve funds transfer

peculiarity:


peculiarity:
Gas calculation using `forge test --match-contract CentriufgeRoutertest --gas-report`
 - Without refactoring:
   **avg**: 224255, **median**:37816, **max**: 500879 , **# calls**": 303
 - With refactoring:
     **avg**: 211735 , **median**: 113595, **max**: 486799 , **# calls**": 303

PR #308, opened 2024-05-25
Authorize operator + renamings to `controller`

hieronx:


src/ERC7540Vault.sol
```diff
@@ -49,6 +51,16 @@ contract ERC7540Vault is Auth, IERC7540 {
     /// @dev    Requests for Centrifuge pool are non-transferable and all have ID = 0
     uint256 constant REQUEST_ID = 0;
 
+    bytes32 private immutable nameHash;
+    bytes32 private immutable versionHash;
+    uint256 public immutable deploymentChainId;
```
peculiarity, line 56:
```
    bytes32 private immutable _NAME_HASH;
    bytes32 private immutable _VERSION_HASH;    
    bytes32 private immutable _DOMAIN_SEPARATOR;    
    uint256 public immutable DEPLOYMENT_CHAIN_ID;
    bytes32 public constant AUTHORIZE_OPERATOR_TYPEHASH = keccak256(
        "AuthorizeOperator(address owner,address operator,bool approved,uint256 validAfter,uint256 validBefore,bytes32 nonce)"
    );    
  ```
  
   Also I'm curious why `AUTHORIZED_OPERATOR_TYPEHASH` is initialized  here and the rest are initialized in the constructor?
   
   Having them initialized in the constructor is cheaper in overall cost.

File: src/interfaces/IERC7540.sol [hunk 2]
```diff
@@ -181,6 +181,18 @@ interface IERC7540CancelRedeem {
         returns (uint256 shares);
 }
 
+interface IAuthorizeOperator {
```
peculiarity, line 184:
```suggestion
interface IAuthorizedOperator {
```

src/ERC7540Vault.sol
```diff
@@ -49,6 +51,16 @@ contract ERC7540Vault is Auth, IERC7540 {
...
+    bytes32 private immutable _DOMAIN_SEPARATOR;
+    bytes32 public constant AUTHORIZE_OPERATOR_TYPEHASH = keccak256(
+        "AuthorizeOperator(address owner,address operator,bool approved,uint256 validAfter,uint256 validBefore,bytes32 nonce)"
+    );
+
+    mapping(address owner => mapping(bytes32 nonce => bool used)) authorizationStates;
```
peculiarity, line 62:
```suggestion
    mapping(address owner => mapping(bytes32 nonce => bool used)) authorizations;
```

test/integration/Operator.t.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity 0.8.21;
 
+import "forge-std/console.sol";
```
peculiarity, line 4: I think we can remove that

src/interfaces/IERC7540.sol
hieronx: The method is `authorizeOperator` so ithink the current name makes sense. But this will be replaced by an ERC number in the future anyways

test/integration/Operator.t.sol
hieronx: Good catch

src/ERC7540Vault.sol
hieronx: No good reason other than copying this over from the erc20 code where it's done similarly.

hieronx: > Having them initialized in the constructor is cheaper in overall cost.

Can you expand on this? Initializing in constructor is cheaper than initializing in the variable declaration?

peculiarity: 1. I think there should be consistency. One is initialized  in the variable declaration an some in the constructor. I think it should be in one place.
2. Naming should be also unified and I see some parts of the solidity style guide were followed and some not. We can agree of course on this :)
3. I was curious why  you would choose constructor initialization over in declaration so I deployed 2 version of the same contract ( on Remix ) .  
  - Version 1: All variables marked as constants and initialized during declaration
  - Version 2: All variables immutable and initialized in the constructor
 Version 2 deployment cost was less than the Version deployment cost.
 
 I don't know why it's that. it's just an observation.

AStox (approved):
Looks good!

AStox (approved):

src/ERC7540Vault.sol
hieronx: Let's put this in a ticket and decide how we want to handle this later

peculiarity (approved):

ilinzweilin (approved):

PR #309, opened 2024-05-27
feat: Introduce cost payments for all components during transaction e…

peculiarity:
…xecution ( bridge + destination chain )


src/gateway/Gateway.sol
```diff
@@ -51,7 +51,15 @@ contract Gateway is Auth, IGateway {
         _;
     }
 
+    event Received(address indexed sender, uint256 amount);
+    
+    // TODO Do we want to have a method that can withdraw the funds from the contract?
```
hieronx, line 56: I think we should add a version of `recoverToken` like this:
```solidity
        if (token === address(0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee ) {
                payable(to).call{value: amount ether}
        } else {
        }
    }
```

The address coming from: https://eips.ethereum.org/EIPS/eip-7528

PR #310, opened 2024-05-27
Remove transfer proxy recovery

hieronx:


ilinzweilin (approved):

AStox (commented):
But isn't this contract meant to make the UX for whoever's executing the transfers better? This PR may reduce code complexity, but I think it's adding complexity for the user now

AStox:
>You can always trigger a Transfer to Centrifuge Chain, and then send it back from there to the recovery address manually.
I think it would be nice to have this illustrated in a test.

AStox (approved):

PR #312, opened 2024-06-06
Automated Gas Payments

peculiarity:


src/CentrifugeGasService.sol
```diff
@@ -0,0 +1,57 @@
...
+        centrifugePrice = centrifugePrice_;
+        updatedTime = block.timestamp;
+        roundId = 1;
+    }
+
+    function file(bytes32 what, uint256 value) external auth {
+        if (what == "centrifugeCost") {
+            centrifugeCost = value;
+        }
+        if (what == "centrifugePrice") {
```
hieronx, line 30: nit:
```suggestion
        if (what == "cost") {
            cost = value;
        }
        if (what == "price") {
```

src/gateway/Gateway.sol
```diff
@@ -72,7 +81,17 @@ contract Gateway is Auth, IGateway {
             msg.sender == investmentManager || msg.sender == poolManager || msg.sender == messages[message.toUint8(0)],
             "Gateway/invalid-manager"
         );
-        aggregator.send(message);
+
+        if (MessagesLib.messageType(message) == MessagesLib.Call.EntryPoint) {
```
hieronx, line 85: I'm confused, what do we actually need `entrypoint` for?

src/gateway/routers/Aggregator.sol
```diff
@@ -189,15 +207,53 @@ contract Aggregator is Auth, IAggregator {
...
+            // Resolution 2: During the TX runtime, we get latest updated value in all our contract for all the fee
+            // and make calculation based on that. If the provide value "fuel" turns out to be not enough, we receive
+            // a faster feedback and we don't have to submit the tx on axelar or any other bridge and wait for a
+            // feedback that
+            // the tx will end up underpaid
+            // Resolution 3: Resolution 2 applies - we get the latest updated costs on centra and axelar assuming they
+            // got updated
+
+            if (fuel > 0) {
+                uint256 txCost = currentRouter.estimate(payload) + centrifugeCost;
```
hieronx, line 242: I would pass `centrifugeCost` to `estimate` and include it in the calculation there. There might be an adapter in the future that actually takes a % of the total cost rather than a fixed fee, and this would then allow implementing that (so the cost of the adapter would be `1.1 * centrifugeCost` as an example)

src/gateway/routers/Aggregator.sol
```diff
@@ -218,4 +274,17 @@ contract Aggregator is Auth, IAggregator {
     function votes(bytes32 messageHash) external view returns (uint16[8] memory) {
         return messages[messageHash].votes;
     }
+
+    /// @inheritdoc IAggregator
+    function estimate(bytes calldata payload) external view returns (uint256 estimated) {
+        bytes memory proof = abi.encodePacked(uint8(MessagesLib.Call.MessageProof), keccak256(payload));
+        uint256 centrifugeExecutionCost = centrifugeGasService.estimate(payload);
+        uint256 numRouters = routers.length;
+        require(numRouters > 0, "Aggregator/not-initialized");
```
hieronx, line 283: I think this is not really required
```suggestion
        uint256 numRouters = routers.length;
```

src/gateway/routers/axelar/Router.sol
```diff
@@ -19,24 +20,49 @@ interface AggregatorLike {
...
 
     AggregatorLike public immutable aggregator;
     AxelarGatewayLike public immutable axelarGateway;
+    AxelarGasServiceLike public immutable axelarGasService;
 
-    constructor(address aggregator_, address axelarGateway_) {
+    // This value is in AXELAR fees in ETH ( wei )
+    uint256 axelarCost = 58039058122843;
+
+    event File(bytes32 what, uint256 value);
```
hieronx, line 49: Should be in the interface `IRouter`

src/gateway/Gateway.sol
peculiarity: The standard flow is
-  _EOA -> ERC7540Vault -> Investment Manager -> Gateway -> ..._

With the UX improvements and CentrifugeRouter introduction the flow is:
- _Protocol -> CentrifugeRouter -> ERC7540Vault -> Investment Manager -> Gateway -> ..._

When the tx is triggered through the CentrifugeRouter, there are funds passed along the way to pay for the tx costs on all adapters + dst chain.

There are couple of scenarios we would like to consider:
 -  The value corresponding to amount that needs to be paid to all bridges + dst chain ( tx cost ) is a pessimistic value. That means that is a majority of the cases it will higher than the actual amount. The remaining of the funds will be sent to the gateway thus the gateway will contain funds. Given this, we don't want users following first flow, take advantage of the remaining gateway funds and pay for them.
 - We would like to pre-pay for an user ( following first flow ) so we will send ETH to the gateway directly. We should allow the tx cost payment for that user.
 
 This is something me and @ilinzweilin  discussed. We agreed that it's important to differentiate the entry point even though it's not clear right now how it will work and what are the cases exactly where we going to need this.
 
 Let me know if you have questions

peculiarity: I think it needs polishing. This is a rough implementation. Frankly I spent significant time how to compromise and do minimal changes to implement this ( use existing functionality ) and not introducing any new methods and code changes.

src/CentrifugeGasService.sol
peculiarity: Agree. In general we have to strive for simplicity in naming. Since there is no specific need to use `centrifuge` to distinguish the variable ( like two price variables or cost variables ) we can just use simple words.

src/CentrifugeRouter.sol
```diff
@@ -7,18 +7,24 @@ import {SafeTransferLib} from "src/libraries/SafeTransferLib.sol";
...
         SafeTransferLib.safeTransfer(token, to, amount);
     }
 
+    // TODO distinguish the source of the TX  whether it comes from CentrifugeRouter or ERC7540Vault directrly.
     // --- Deposit ---
-    /// @inheritdoc ICentrifugeRouter
-    function requestDeposit(address vault, uint256 amount) external {
+    function requestDeposit(address vault, uint256 amount) external payable {
+        require(msg.value > 0, "CentrifugeRouter/not-enough-gas-funds");
+        payable(gateway).transfer(msg.value);
```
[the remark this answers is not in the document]
peculiarity: This might change the whole idea around the entry point.

src/gateway/routers/Aggregator.sol
peculiarity: Just for the protocol: I don't feel like passing centrifuge cost  to the estimate method is the right way. IMO the `GasService` should be a dependency  in every adapter.  It is an essential part of it. If we use `adapter.estimate` in other contracts we have to inject `GasService` there as well only to get the `centrifugeCost`  and pass it to the `estimate` method.
The only reason why I also agree with this approach is the fact that if we put `GasService` as a dep on each router, there will be calls to the gas service as many times as the amount of routers instead of having it just once in the Aggregator.
I will make the change I just wanted to share my though.

src/gateway/routers/Aggregator.sol
```diff
@@ -218,4 +278,14 @@ contract Aggregator is Auth, IAggregator {
     function votes(bytes32 messageHash) external view returns (uint16[8] memory) {
         return messages[messageHash].votes;
     }
+
+    /// @inheritdoc IAggregator
+    function estimate(bytes calldata payload) external view returns (uint256 estimated) {
```
hieronx, line 283: Wdyt about moving this to `Gateway` instead? It seems to fit more intuitively there. Would be a bit more complex to get the routers but should be possible.

src/gateway/Gateway.sol
```diff
@@ -51,7 +54,13 @@ contract Gateway is Auth, IGateway {
         _;
     }
 
+    event Received(address indexed sender, uint256 amount);
+
+    receive() external payable {
+        emit Received(msg.sender, msg.value);
+    }
     // --- Administration ---
+
```
hieronx, line 63:
```suggestion
```

src/gateway/Gateway.sol
```diff
@@ -51,7 +54,13 @@ contract Gateway is Auth, IGateway {
         _;
     }
 
+    event Received(address indexed sender, uint256 amount);
```
hieronx, line 57: Should be moved to the interface

src/gateway/GasService.sol
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    function file(bytes32 what, uint256 value) external auth {
+        if (what == "messageCost") {
+            messageCost = value;
+        }
+        if (what == "proofCost") {
+            proofCost = value;
+        }
+        if (what == "price") {
```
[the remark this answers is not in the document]
peculiarity: Do you are suggesting to use `updatePrice` method ?

src/gateway/routers/axelar/Router.sol
```diff
@@ -19,24 +20,47 @@ interface AggregatorLike {
...
     bytes32 public constant CENTRIFUGE_ID_HASH = keccak256(bytes("centrifuge"));
     bytes32 public constant CENTRIFUGE_ADDRESS_HASH = keccak256(bytes("0x7369626CEF070000000000000000000000000000"));
     string public constant CENTRIFUGE_AXELAR_EXECUTABLE = "0xc1757c6A0563E37048869A342dF0651b9F267e41";
 
     AggregatorLike public immutable aggregator;
     AxelarGatewayLike public immutable axelarGateway;
+    AxelarGasServiceLike public immutable axelarGasService;
 
-    constructor(address aggregator_, address axelarGateway_) {
+    /// @dev This value is in AXELAR fees in ETH ( wei )
```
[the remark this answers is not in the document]
peculiarity: Um.... this is a variable

src/gateway/routers/Aggregator.sol
peculiarity: Main reason why I left everything in the aggregator is because routers are there.
Getting the routers won't be an issue since they are a public variable.
What bothers me is that in order to make this work I have to also provide `GasService` here as well.
You basically get duplication of "functionality". Why not just keep it in the aggregator?

On second thought... Since this will be a view function, what stops us to create smth like this.

```solidity
contract Gateway {

 //......
 
   function estimate() external view returns (uint256) {
       return AggregatorLike(aggregator).estimate();
    }
 }
 ```

Comment (hieronx) [reply src/gateway/routers/Aggregator.sol:280]: Maybe we should actually keep the `estimate` implementation in aggregator, and add a view method like the above to the `CentrifugeRouter`. So that is a "one stop shop" for interacting with Centrifuge

File: src/gateway/routers/axelar/Router.sol (hunk 11 above)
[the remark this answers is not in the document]
Comment (hieronx) [reply src/gateway/routers/axelar/Router.sol:46]: Yeah fair point but that is actually something that we still need to decide. Quite a few others are using the same interface pattern also for variables. See e.g. https://github.com/morpho-org/morpho-blue/blob/main/src/Morpho.sol#L53

I'm a bit 50/50 on whether we should also adopt this

File: src/gateway/routers/Aggregator.sol [hunk 12]
```diff
@@ -189,15 +204,57 @@ contract Aggregator is Auth, IAggregator {
...
+        if (fuel > 0) {
+            destChainMsgCost = gasService.estimate(message);
+            destChainProofCost = gasService.estimate(proof);
+        }
+
         for (uint256 i; i < numRouters; i++) {
-            RouterLike(routers[i]).send(i == PRIMARY_ROUTER_ID - 1 ? message : proof);
+            RouterLike currentRouter = RouterLike(routers[i]);
+            bytes memory payload = i == PRIMARY_ROUTER_ID - 1 ? message : proof;
+            // TODO a discussion:
```
peculiarity, line 226: @hieronx  have you thought about this one?
If you agree we it , let me know so that I can remove this comment

src/gateway/routers/Aggregator.sol
peculiarity: This second suggestion is actually better. Indeed we would like to make this abstracted as much as possible and make the integrator use only `CentrifugeRouter`.

src/gateway/routers/axelar/Router.sol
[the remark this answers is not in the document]
peculiarity: Well I can argue about this specific case.
Owner is a generic variable that is used in most of the contract. It is not protocol / implementation specific .
In our case this is a very Axelar specific variable. It doesn't live outside this implementation.

Even though this variable per se is axelar specific, what we could do is actually make it more generic and add the `getter` as part of the interface ( What's been done in morpho ). In Axelar it's a constant but in other implementations it could be a calculated value.

src/gateway/routers/Aggregator.sol
hieronx: Yeah I was still thinking about this a bit more, but for now I think the solution you have gone for makes sense.

src/gateway/GasService.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    function file(bytes32 what, uint256 value) external auth {
+        if (what == "messageCost") messageCost = value;
+        if (what == "proofCost") proofCost = value;
+        else revert("CentrifugeGasService/file-unrecognized-param");
+        emit File(what, value);
+    }
+
+    function updatePrice(uint256 value) external auth {
```
hieronx, line 39: I think we should consider actually adding a message for this. So Centrifuge Chain can periodically send messages updating the gas price directly.

src/gateway/GasService.sol
```diff
@@ -0,0 +1,60 @@
...
+        price = value;
+        roundId++;
+        updatedTime = block.timestamp;
+    }
+
+    function estimate(bytes calldata payload) public view returns (uint256) {
+        if (MessagesLib.messageType(payload) == MessagesLib.Call.MessageProof) {
+            return proofCost.mulDiv(price, 10 ** 18);
```
hieronx, line 47: I think it's better to always the version with explicit rounding.
```suggestion
            return proofCost.mulDiv(price, 10 ** 18, MathLib.Rounding.Up);
```

src/gateway/GasService.sol
```diff
@@ -0,0 +1,60 @@
...
+        }
+        return messageCost.mulDiv(price, 10 ** 18);
```
hieronx, line 49:
```suggestion
        return messageCost.mulDiv(price, 10 ** 18, MathLib.Rounding.Up);
```

src/gateway/Gateway.sol
```diff
@@ -89,11 +106,24 @@ contract Gateway is Auth, IGateway {
             manager = address(root);
         } else {
             // Dynamic path for other managers, to be able to easily
-            // extend functionality of Liquidity Pools
+            // extend functionality of Liquidity Poolsif(msg.sender == address(mana))
```
hieronx, line 109: Why this comment change?

peculiarity: Because I have fat fingers and I might have cp'ed something at that line :rofl:

src/gateway/GasService.sol
peculiarity: Adding a message as a message within the MessageLib and handle it ?

src/gateway/Gateway.sol
```diff
@@ -72,7 +79,17 @@ contract Gateway is Auth, IGateway {
             msg.sender == investmentManager || msg.sender == poolManager || msg.sender == messages[message.toUint8(0)],
             "Gateway/invalid-manager"
         );
-        aggregator.send(message);
+
+        if (MessagesLib.messageType(message) == MessagesLib.Call.EntryPoint) {
```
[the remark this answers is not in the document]
ilinzweilin: in that way we would be more flexible and not depend on the source, but rather care about the fact if the user should have prepaid the tx or not

src/PoolManager.sol
```diff
@@ -464,6 +473,12 @@ contract PoolManager is Auth, IPoolManager {
         emit RemoveVault(poolId, trancheId, asset, vault);
     }
 
+    function updateDestChainTokenPrice(uint256 price) public auth {
```
hieronx, line 476: I think this is more descriptive:
```suggestion
    function updateCentrifugeGasPrice(uint256 price) public auth {
```

hieronx: (and would update the message etc accordingly)

src/gateway/GasService.sol
```diff
@@ -0,0 +1,53 @@
...
+    using MathLib for uint256;
+
+    /// @inheritdoc IGasService
+    uint256 public price;
+    /// @inheritdoc IGasService
+    uint256 public messageCost;
+    /// @inheritdoc IGasService
+    uint256 public proofCost;
+
+    constructor(uint256 messageCost_, uint256 proofCost_, uint256 price_) {
```
hieronx, line 19: nit: would put `price` first since it's also first in the variable list above. Also personally I think aesthetically it is always nice to order arguments by shortest to longest (if there is no other semantic reason for the ordering), but that is a super nit ;)

src/gateway/GasService.sol
```diff
@@ -0,0 +1,53 @@
...
+        messageCost = messageCost_;
+        proofCost = proofCost_;
+        price = price_;
+
+        wards[msg.sender] = 1;
+        emit Rely(msg.sender);
+    }
+    /// @inheritdoc IGasService
+
```
hieronx, line 28:
```suggestion

    /// @inheritdoc IGasService
```

src/gateway/GatewayV2.sol
```diff
@@ -0,0 +1,343 @@
...
+    address public poolManager;
+    GasServiceLike public gasService;
+
+    address[] public routers;
+    mapping(address router => Router) public activeRouters;
+    mapping(bytes32 messageHash => Message) public messages;
+    mapping(bytes32 messageHash => Recovery) public recoveries;
+    mapping(uint8 messageId => address manager) messageHandlers;
+
+    uint256 allowance;
```
hieronx, line 59: I'm not a huge fan of the name allowance since it overlaps with ERC20 allowance/approvals. What about `quota`? Or even just `gasAllowance`

src/gateway/GatewayV2.sol
```diff
@@ -0,0 +1,343 @@
...
+        require(recovery.timestamp <= block.timestamp, "Gateway/challenge-period-has-not-ended");
+        require(router.id != 0, "Gateway/invalid-router");
+
+        delete recoveries[messageHash];
+        _handle(message, recovery.router, router, true);
+        emit ExecuteMessageRecovery(message);
+    }
+
+    // --- Outgoing ---
+
```
hieronx, line 242:
```suggestion
```

src/gateway/GatewayV2.sol
```diff
@@ -0,0 +1,343 @@
...
+                remaining = tank - consumed;
+            }
+            require(remaining < tank, "Gateway/not-enough-gas-funds");
+            tank = remaining;
+
+            currentRouter.pay{value: consumed}(payload, address(this));
+            currentRouter.send(payload);
+        }
+
+        if (fuel > 0 && tank > 0) allowance = 0;
```
hieronx, line 272: Looks to me like the separate variable is not really needed here.
```suggestion
        uint256 tank = allowance > 0 ? allowance : gasService.shouldRefuel(source, message) ? address(this).balance : 0;

        uint256 destChainMsgCost = gasService.estimate(message);
        uint256 destChainProofCost = gasService.estimate(proof);

        for (uint256 i; i < numRouters; i++) {
            RouterLike currentRouter = RouterLike(routers[i]);
            bool isPrimaryRouter = i == PRIMARY_ROUTER_ID - 1;
            bytes memory payload = isPrimaryRouter ? message : proof;

            uint256 consumed = currentRouter.estimate(payload, isPrimaryRouter ? destChainMsgCost : destChainProofCost);
            uint256 remaining;
            unchecked {
                remaining = tank - consumed;
            }
            require(remaining < tank, "Gateway/not-enough-gas-funds");
            tank = remaining;

            currentRouter.pay{value: consumed}(payload, address(this));
            currentRouter.send(payload);
        }

        if (allowance > 0 && tank > 0) allowance = 0;
```

src/gateway/routers/axelar/Router.sol
```diff
@@ -60,4 +84,16 @@ contract AxelarRouter is IRouter {
...
+    function pay(bytes calldata payload, address refund) public payable {
+        axelarGasService.payNativeGasForContractCall{value: msg.value}(
+            address(this), CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, payload, refund
+        );
+    }
+
+    /// @dev Currently the payload ( message ) is not taken into consideration during cost estimation
+    /// A predefined `axelarCost` value is used.
+    function estimate(bytes calldata, uint256 destChainCost) public view returns (uint256) {
+        return axelarCost + destChainCost;
```
hieronx, line 97: I think we can just call this `gasLimit`
```suggestion
    function estimate(bytes calldata, uint256 gasLimit) public view returns (uint256) {
        return axelarCost + gasLimit;
```

src/gateway/GatewayV2.sol
hieronx: I think `destChainMsgCost` and `destChainProofCost` here can just be `messageCost` and `proofCost`

src/gateway/GatewayV2.sol
peculiarity: I like `quota`. I also had some doubts about the allowance

src/PoolManager.sol
peculiarity: I wanted to make it more generic. Not sure if it's needed thought.  If you are ok with having it specific to centrifuge, then it's fine with me

src/gateway/GatewayV2.sol
peculiarity: Variables  changed.
`fued` variable is used to cache the storage one in order to save gas from reads.

src/PoolManager.sol
```diff
@@ -200,6 +211,8 @@ contract PoolManager is Auth, IPoolManager {
             unfreeze(message.toUint64(1), message.toBytes16(9), message.toAddress(25));
         } else if (call == MessagesLib.Call.DisallowAsset) {
             disallowAsset(message.toUint64(1), message.toUint128(9));
+        } else if (call == MessagesLib.Call.UpdateCentrifugeGasPrice) {
+            updateCentrifugeGasPrice(message.toUint256(1));
```
ilinzweilin, line 215: shouldn't we use`uint128` here, since Centrifuge chain supports `uint128`

src/PoolManager.sol
```diff
@@ -463,6 +476,12 @@ contract PoolManager is Auth, IPoolManager {
         emit RemoveVault(poolId, trancheId, asset, vault);
     }
 
+    function updateCentrifugeGasPrice(uint256 price) public auth {
```
ilinzweilin, line 479: I think price should also be `uint128`

src/PoolManager.sol
```diff
@@ -30,6 +30,11 @@ interface AuthLike {
     function deny(address user) external;
 }
 
+interface GasServiceLike {
+    function updatePrice(uint256 value) external;
```
ilinzweilin, line 34: I think the function is called `updateGasPrice` & it should be a `uint128`

src/PoolManager.sol
```diff
@@ -463,6 +476,12 @@ contract PoolManager is Auth, IPoolManager {
         emit RemoveVault(poolId, trancheId, asset, vault);
     }
 
+    function updateCentrifugeGasPrice(uint256 price) public auth {
+        require(price > 0, "PoolManager/price-cannot-be-zero");
+        require(gasService.price() != price, "PoolManager/same-price-already-set");
+        gasService.updatePrice(price);
```
ilinzweilin, line 482: `updateGasPrice`

src/gateway/GasService.sol
```diff
@@ -0,0 +1,67 @@
...
+    /// @inheritdoc IGasService
+    function file(bytes32 what, uint64 value) external auth {
+        if (what == "messageCost") messageCost = value;
+        if (what == "proofCost") proofCost = value;
+        else revert("CentrifugeGasService/file-unrecognized-param");
+        emit File(what, value);
+    }
+
+    function updateGasPrice(uint64 value) external auth {
```
ilinzweilin, line 40: gasPrice should be `uint128`

src/gateway/GasService.sol
```diff
@@ -0,0 +1,67 @@
...
+    /// @inheritdoc IGasService
+    uint64 public proofCost;
+    /// @inheritdoc IGasService
+    uint64 public messageCost;
+    /// @inheritdoc IGasService
+    uint64 public gasPrice;
+    /// @inheritdoc IGasService
+    uint256 public tokenPrice;
+
+    constructor(uint64 proofCost_, uint64 messageCost_, uint64 gasPrice_, uint256 tokenPrice_) {
```
ilinzweilin, line 22: why are the proof and messageCost values `uint64`, they can be `uint256`

src/CentrifugeRouter.sol
```diff
@@ -176,6 +202,10 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         return IPoolManager(poolManager).getVault(poolId, trancheId, asset);
     }
 
+    function estimate(bytes calldata payload) external returns (uint256 amount) {
```
ilinzweilin, line 205: I don't think we need to expose that function here. Users can use the gateway directly

src/gateway/Gateway.sol
```diff
@@ -51,47 +73,277 @@ contract Gateway is Auth, IGateway {
...
+    /// @inheritdoc IGateway
     function file(bytes32 what, uint8 data1, address data2) public auth {
-        if (what == "message") messages[data1] = data2;
+        if (what == "message") messageHandlers[data1] = data2;
         else revert("Gateway/file-unrecognized-param");
         emit File(what, data1, data2);
     }
 
+    // --- Incoming ---
+    /// @inheritdoc IGateway
```
ilinzweilin, line 133: handle has to be `auth` 🙈

src/gateway/Gateway.sol
```diff
@@ -51,47 +73,277 @@ contract Gateway is Auth, IGateway {
...
+
+        emit SendMessage(message);
     }
 
-    // --- Incoming ---
-    function handle(bytes calldata message) external auth pauseable {
+    function topUp() external payable {
+        require(RootLike(root).endorsed(msg.sender), "Gateway/only-endorsed-can-topup");
+        require(msg.value > 0, "Gateway/cannot-topup-with-nothing");
+        quota = msg.value;
```
ilinzweilin, line 291: why do we need the extra `quota` variable, would just use `fuel `

src/gateway/Gateway.sol
```diff
@@ -51,47 +73,277 @@ contract Gateway is Auth, IGateway {
...
+        uint256 messageCost = gasService.estimate(message);
+        uint256 proofCost = gasService.estimate(proof);
+
+        for (uint256 i; i < numRouters; i++) {
+            RouterLike currentRouter = RouterLike(routers[i]);
+            bool isPrimaryRouter = i == PRIMARY_ROUTER_ID - 1;
+            bytes memory payload = isPrimaryRouter ? message : proof;
+
+            uint256 consumed = currentRouter.estimate(payload, isPrimaryRouter ? messageCost : proofCost);
+            uint256 remaining;
```
ilinzweilin, line 272: I would not introduce `remaining` as a new var. You are risking an overflow, as tank might be smaller than consumed. Would need to use an int. But also the var is not required. Would do following:

```
   require(consumed < tank, "Gateway/not-enough-gas-funds");
   tank = tank - consumed;
```

src/gateway/Gateway.sol
```diff
@@ -51,47 +73,277 @@ contract Gateway is Auth, IGateway {
...
+                remaining = tank - consumed;
+            }
+            require(remaining < tank, "Gateway/not-enough-gas-funds");
+            tank = remaining;
+
+            currentRouter.pay{value: consumed}(payload, address(this));
+            currentRouter.send(payload);
+        }
+
+        if (fuel > 0 && tank > 0) quota = 0;
```
ilinzweilin, line 283: I would get rid of quota, add fuel to contract state and set it to 0

src/gateway/routers/Aggregator.sol
```diff
@@ -218,4 +275,14 @@ contract Aggregator is Auth, IAggregator {
...
         return messages[messageHash].votes;
     }
+
+    /// @inheritdoc IAggregator
+    function estimate(bytes calldata payload) external view returns (uint256 estimated) {
+        bytes memory proof = abi.encodePacked(uint8(MessagesLib.Call.MessageProof), keccak256(payload));
+        uint256 destChainCost = gasService.estimate(payload);
+
+        for (uint256 i; i < routers.length; i++) {
+            estimated += RouterLike(routers[i]).estimate(i == PRIMARY_ROUTER_ID - 1 ? payload : proof, destChainCost);
```
ilinzweilin, line 285: would not intro `destChainCost` as a var since `gasService.estimate(payload)` is only called once within the function

ilinzweilin (changes requested):

src/gateway/GasService.sol
peculiarity: This is something we agreed with Jeroen.

src/CentrifugeRouter.sol
peculiarity: We also discussed that. Since we would like to have the CentifugeRouter as a one-stop shop, we would like for the integrators / users to care only about this contract and nothing else.

src/gateway/Gateway.sol
peculiarity: I think we agreed that it's fine right now. Only routers can call handle. Correct?

src/gateway/Gateway.sol
peculiarity: I did some renaming. `quota` is updated only by `CentrifugeRouter`. If there is no `quote` assigned of the transaction, then we use the gateway's balance .

src/gateway/Gateway.sol
hieronx: Yeah I agree this is correct as is.

src/gateway/Gateway.sol
peculiarity: Actually I was relying on the underflow. That's why there was this `unchecked`. But I agree that we don't need the remaining. 
Cheers.

src/gateway/Gateway.sol
peculiarity: No matter how to name the state variable, there must be a local memory cache variable to save on SLOAD gas.

src/gateway/routers/Aggregator.sol
peculiarity: This is aggregator :D This will be deleted

hieronx (approved):

script/Deployer.sol
```diff
@@ -58,8 +57,13 @@ contract Deployer is Script {
         investmentManager = new InvestmentManager(address(root), address(escrow));
         poolManager = new PoolManager(address(escrow), vaultFactory, restrictionManagerFactory, trancheTokenFactory);
 
+        // TODO THESE VALUES NEEDS TO BE CHECKED
+        gasService = new GasService(20000000000000000, 20000000000000000, 2500000000000000000, 178947400000000);
```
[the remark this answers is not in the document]
peculiarity: I think that once we come up wuth the actual values, we can put the into env vars :D

ilinzweilin (approved):

PR #315, opened 2024-06-10
feat: Check if source of the transaction is an endorsed address and i…

peculiarity:
…f so, pay for the transaction

### DIFF
--- a/src/gateway/Gateway.sol
+++ b/src/gateway/Gateway.sol
@@ -33,7 +33,6 @@ contract Gateway is Auth, IGateway {
 
     RootLike public immutable root;
 
-    address entryPoint;
     address public poolManager;
     address public investmentManager;
     AggregatorLike public aggregator;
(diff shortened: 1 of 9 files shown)


src/gateway/Gateway.sol
```diff
@@ -74,22 +73,14 @@ contract Gateway is Auth, IGateway {
...
-            entryPoint = message.toAddress(1);
-            return;
-        }
-
-        aggregator.send{value: address(this).balance}(message);
-
-        if (entryPoint != address(0)) {
-            delete entryPoint;
-        }
+        uint256 gasFunds = isPrepaid ? address(this).balance : 0;
```
ilinzweilin, line 82: I don't think that this line is correct as in both cases you would use the balance, but for prepaid tx we somehow need to make sure that the user sent enough funds.

So sth like: 
`uint256 gasFunds = isPrepaid ? userGasBalance : address(this).balance;`

ilinzweilin (changes requested):

PR #318, opened 2024-06-12
Unite Gateway and Aggregator in a single contract.

peculiarity:


src/CentrifugeRouter.sol
```diff
@@ -45,11 +50,13 @@ contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
...
 
     // --- Deposit ---
-    function requestDeposit(address vault, uint256 amount) external payable {
-        require(msg.value > 0, "CentrifugeRouter/not-enough-gas-funds");
-        // Due to the multicall nature of this contract, there are issues calling `transfer` multiple times.
-        // Investigate further.
-        payable(gateway).call{value: msg.value}("");
+    function requestDeposit(address vault, uint256 amount, uint256 topUpAmount) external payable {
+        // Assumption: This contract won't hold any ETH
+        require(topUpAmount <= msg.value, "CentrifugeRouter/must-pay-for-tx-costs");
```
ilinzweilin, line 55: if we are not using the real estimation from gateway, the user can just send a random topup value.

src/CentrifugeRouter.sol
```diff
@@ -9,6 +9,11 @@ import {IERC20} from "src/interfaces/IERC20.sol";
 import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
 import {IPoolManager} from "src/interfaces/IPoolManager.sol";
 
+interface GatewayLike {
+    function topUp() external payable;
```
ilinzweilin, line 13: I would call topup prepay & use the balance in case the tx is coming from a prepaid source

src/CentrifugeRouter.sol
```diff
@@ -45,11 +50,13 @@ contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
...
-        require(msg.value > 0, "CentrifugeRouter/not-enough-gas-funds");
-        // Due to the multicall nature of this contract, there are issues calling `transfer` multiple times.
-        // Investigate further.
-        payable(gateway).call{value: msg.value}("");
+    function requestDeposit(address vault, uint256 amount, uint256 topUpAmount) external payable {
+        // Assumption: This contract won't hold any ETH
+        require(topUpAmount <= msg.value, "CentrifugeRouter/must-pay-for-tx-costs");
+        // In a situation where we have multiple calls ( as in multicall ) the msg.value remains the same
+        // but with each call the balance will get reduced by `topUp`
+        require(topUpAmount <= address(this).balance, "CentrifugeRouter/insufficient-funds-to-topup");
```
ilinzweilin, line 58: I don't understand this check. Even if you call `requestDeposit` multiple times within 1 tx, you would still need gas for multiple tx to send multiple requests to cfg chain

src/gateway/GatewayV2.sol
```diff
@@ -0,0 +1,347 @@
...
+    function estimate(bytes calldata payload) external view returns (uint256[] memory tranches, uint256 total) {
+        bytes memory proof = abi.encodePacked(uint8(MessagesLib.Call.MessageProof), keccak256(payload));
+        uint256 destChainCost = gasService.estimate(payload);
+        tranches = new uint256[](routers.length);
+
+        for (uint256 i; i < routers.length; i++) {
+            uint256 estimated =
+                RouterLike(routers[i]).estimate(i == PRIMARY_ROUTER_ID - 1 ? payload : proof, destChainCost);
+            tranches[i] = estimated;
+            total += estimated;
```
ilinzweilin, line 315: I was thinking that these variables can be stored as state variables & set during gas estimation check in the cfg router

src/gateway/GatewayV2.sol
```diff
@@ -0,0 +1,347 @@
...
+        uint256 numRouters = routers.length;
+        require(numRouters > 0, "Gateway/not-initialized");
+        uint256 tank = fuel;
+
+        uint256 destChainMsgCost;
+        uint256 destChainProofCost;
+
+        bytes memory proof = abi.encodePacked(uint8(MessagesLib.Call.MessageProof), keccak256(message));
+
+        if (tank > 0) {
```
ilinzweilin, line 253: there might be enough funds in the gateway for tx that didn't require funding from users (= don't need to be actively prepaid / call did not come from centrifuge router). So tank in that case can still be 0 and the tx should go through and be funded through funds held in gateway

src/CentrifugeRouter.sol
```diff
@@ -9,6 +9,11 @@ import {IERC20} from "src/interfaces/IERC20.sol";
 import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
 import {IPoolManager} from "src/interfaces/IPoolManager.sol";
 
+interface GatewayLike {
+    function topUp() external payable;
+    function estimate(bytes calldata payload) external returns (uint256[] memory tranches, uint256 total);
```
ilinzweilin, line 14: I would combine topup and estimate to one function call that:
- returns the gas estimation
- sets the prepaid amount / fuel balance for prepaid tx

ilinzweilin (changes requested):

src/CentrifugeRouter.sol
peculiarity: I don't it makes much difference whether they sent  less that it is needed. If it's less than the calculated value then the transaction will fail and they just wasted funds on a failed transaction.

Let me tell you exactly what's happening here. This applies  when there is a multicall transaction.

When the user calls `CentrifugeRouter.multicall([requestDeposit,requestDepost, requestDeposit])`  they send one `msg.value` which should be enough to cover for all three requests. The `topUp` amount is just a fraction of the total amount. You don't want to send all of the `msg.value` to the gateway initially because on the second call the balance of the CentrifugeRouter is `0` even though that `msg.value` is 300 ( 3 x 100 ) for example. It causes issues. 
Let me know if I need to clarify further.

src/CentrifugeRouter.sol
peculiarity: Check my comment for the previous line.

src/gateway/GatewayV2.sol
peculiarity: Here is the reasoning why I think it's not ok to estimate from the CfgRouter itself.
The estimation is based on the type of the message ( message or proof ) but not the message itself.
I can easily call `.estimate()` and it will return some values.  That approach is not very future proof and if for some reason the message is going to be needed then we will have to construct the message ( as in the `InvestmentManager` ) and pass it to the estimate method.

src/CentrifugeRouter.sol
ilinzweilin: ok, cool.
I still think we should work with the exact gas estimation

src/gateway/GatewayV2.sol
ilinzweilin: can't we just call the existing estimate function from the centrifuge router and store `(uint256[] memory tranches, uint256 total)` as state variables in the gatewayV2

src/interfaces/IInvestmentManager.sol
```diff
@@ -45,7 +45,7 @@ interface IInvestmentManager {
     ///         proceed with tranche token payouts in case their orders got fulfilled.
     /// @dev    The user asset amount required to fulfill the deposit request have to be locked,
     ///         even though the tranche token payout can only happen after epoch execution.
-    function requestDeposit(address vault, uint256 assets, address receiver, address owner, address souce)
+    function requestDeposit(address vault, uint256 assets, address receiver, address owner, address source)
```
peculiarity, line 48: I think this is something we are interested in general.
I noticed that in the events ( such as  the Deposit one ) we emit who is the sender. 
Given that, I have decided that we can added anyways and use it.

src/InvestmentManager.sol
```diff
@@ -103,16 +105,17 @@ contract InvestmentManager is Auth, IInvestmentManager {
...
-            abi.encodePacked(
-                uint8(MessagesLib.Call.IncreaseInvestOrder),
-                poolId,
-                vault_.trancheId(),
-                receiver,
-                poolManager.assetToId(asset),
-                _assets,
-                source
+            abi.encode(
+                source,
```
peculiarity, line 110: I have decided  to  have something like 
```
  (
        address source; // metadata property
        .... // some other meadata property
       bytes message;      
  )
```

`source` is not part of the message. I don't think that this is needed  on the centrifuge side. This is something we care on our side. Also it allows us to extent this metadata structure without affecting the size of the message.

peculiarity: Choose `encode` instead of `encodePacked` because I think it will be easier to decode instead of parsing byte by byte. We can use ` encodePacked` and in the `send` just get the first 20 bytes for the address and the remaining for the message.

PR #319, opened 2024-06-12
Support multiple parallel recoveries

hieronx:


src/gateway/routers/Aggregator.sol
```diff
@@ -34,7 +34,7 @@ contract Aggregator is Auth, IAggregator {
     address[] public routers;
     mapping(address router => Router) public activeRouters;
     mapping(bytes32 messageHash => Message) public messages;
-    mapping(bytes32 messageHash => Recovery) public recoveries;
+    mapping(address router => mapping(bytes32 messageHash => uint256 timestamp)) public recoveries;
```
hieronx, line 37: Since the router is part of the mapping now, the value doesn't need to be a struct anymore

src/gateway/routers/Aggregator.sol
```diff
@@ -82,12 +82,13 @@ contract Aggregator is Auth, IAggregator {
...
     function handle(bytes calldata payload) external {
-        Router memory router = activeRouters[msg.sender];
-        require(router.id != 0, "Aggregator/invalid-router");
-        _handle(payload, msg.sender, router, false);
+        _handle(payload, msg.sender, false);
     }
 
-    function _handle(bytes calldata payload, address routerAddr, Router memory router, bool isRecovery) internal {
+    function _handle(bytes calldata payload, address router_, bool isRecovery) internal {
+        Router memory router = activeRouters[router_];
```
hieronx, line 89: Realized we had this check in `handle` and `executeMessageRecovery`, both of which call `_handle`, so we could just move it

ilinzweilin (approved):

peculiarity (approved):

PR #320, opened 2024-06-13
ERC20 wrapping and misc `CentrifugeRouter` improvements

hieronx:


src/CentrifugeRouter.sol
```diff
@@ -64,30 +60,90 @@ contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
...
+
+        amount = MathLib.min(amount, IERC20(underlying).balanceOf(msg.sender));
+        require(amount != 0, "CentrifugeRouter/zero-balance");
+        SafeTransferLib.safeTransferFrom(underlying, msg.sender, address(this), amount);
+
+        require(IERC20Wrapper(wrapper).depositFor(receiver, amount), "CentrifugeRouter/deposit-for-failed");
+    }
+
+    function unwrap(address wrapper, uint256 amount, address receiver) external {
+        require(receiver != address(0), "CentrifugeRouter/zero-address");
```
peculiarity, line 116: Does this add the necessary security ? There was this notion that we can still pass wrong address

src/CentrifugeRouter.sol
```diff
@@ -64,30 +60,90 @@ contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
...
+        amount = MathLib.min(amount, IERC20(wrapper).balanceOf(address(this)));
```
peculiarity, line 117: Is this  good UX ? I ask the contract to give me 100 units but it sends back to me 50 only because it's all it had? If I have no underlying knowledge about the implementation I would be frightened.

src/CentrifugeRouter.sol
```diff
@@ -64,30 +60,90 @@ contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
...
+    function permit(address asset, uint256 assets, uint256 deadline, uint8 v, bytes32 r, bytes32 s) external {
+        try IERC20Permit(asset).permit(msg.sender, address(this), assets, deadline, v, r, s) {} catch {}
+    }
+
+    // --- ERC20 wrapping ---
+    /// @dev requires calling approveMax(underlying, wrapper) before
+    function wrap(address wrapper, uint256 amount, address receiver) external {
+        address underlying = IERC20Wrapper(wrapper).underlying();
+
+        amount = MathLib.min(amount, IERC20(underlying).balanceOf(msg.sender));
```
peculiarity, line 108: Again IMO bad UX

src/CentrifugeRouter.sol
hieronx: I basically copied this from Morpho: https://github.com/morpho-org/morpho-blue-bundlers/blob/main/src/ERC20WrapperBundler.sol#L33

But I see your point, let me think about it!

hieronx: I realized why it is the way it is, and it makes sense: if you are chaining multiple actions, you might not know the exact output of some of the actions. So with this, you can overestimate the output of 1 action and provide that as the input to a later action. That wouldn't work if you force exact amounts.

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,186 @@
...
+interface EscrowLike {
+    function approve(address token, address spender, uint256 value) external;
+}
+
+contract CentrifugeRouter is Auth, ICentrifugeRouter {
+    address public immutable escrow;
+
     address public poolManager;
 
+    address constant UNSET_INITIATOR = address(1);
```
hieronx, line 21: This reentrancy lock is inspired by https://github.com/morpho-org/morpho-blue-bundlers/blob/main/src/BaseBundler.sol#L25

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,186 @@
...
         wards[msg.sender] = 1;
         emit Rely(msg.sender);
     }
 
-    // --- Administration ---
-    /// @inheritdoc ICentrifugeRouter
-    function recoverTokens(address token, address to, uint256 amount) external auth {
-        SafeTransferLib.safeTransfer(token, to, amount);
+    modifier protected() {
+        if (_initiator == UNSET_INITIATOR) {
```
hieronx, line 36: We want to allow users to not go through multicall. In that case the `_initiator` should be set here. Otherwise it is set in `multicall`

script/Deployer.sol
```diff
@@ -57,7 +58,9 @@ contract Deployer is Script {
         investmentManager = new InvestmentManager(address(root), address(escrow));
         poolManager = new PoolManager(address(escrow), vaultFactory, restrictionManagerFactory, trancheTokenFactory);
 
-        centrifugeRouter = new CentrifugeRouter(address(poolManager));
+        routerEscrow = new Escrow(deployer);
```
hieronx, line 61: Custom escrow because
1. Cleaner to not mix funds with core contracts
2. Avoids the vault doing a transfer from escrow to escrow

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,186 @@
...
 import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
 import {IPoolManager} from "src/interfaces/IPoolManager.sol";
 
-contract CentrifugeRouter is Auth, Multicall, ICentrifugeRouter {
+interface EscrowLike {
+    function approve(address token, address spender, uint256 value) external;
+}
+
+contract CentrifugeRouter is Auth, ICentrifugeRouter {
+    address public immutable escrow;
```
hieronx, line 17: Escrow is required because intended usage of the Router allows moving funds out of it (using `_approveMax` in e.g. `wrap`), so we have to separate the locked request funds to not expose them.

src/CentrifugeRouter.sol
hieronx: I think we can possibly even remove this require, it doesn't seem to add any security. Really just adapted from Morpho but I can't think of a good reason (besides an auditor commenting that there should be address zero checks..)

src/CentrifugeRouter.sol
peculiarity: Indeed it makes sense the more you think about it :)

src/token/ERC20.sol
```diff
@@ -131,7 +131,7 @@ contract ERC20 is Auth, IERC20Metadata, IERC20Permit {
         emit Transfer(address(0), to, value);
     }
 
-    function burn(address from, uint256 value) external auth {
+    function burn(address from, uint256 value) public auth {
```
peculiarity, line 134: Any particular reason why this is changed to public ?

test/mocks/SimpleERC20Wrapper.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.21;
+
+import {ERC20} from "src/token/ERC20.sol";
+import {IERC20Wrapper, IERC20Metadata} from "src/interfaces/IERC20.sol";
+import "forge-std/Test.sol";
+import {console} from "forge-std/console.sol";
```
peculiarity, line 7: This is not needed. Also AFAIK you don't have to import this explicitly. It's part of the `Test.sol` anyways.

src/token/ERC20.sol
hieronx: No, good catch. Was to be able to wrap it in the `SimpleERC20Wrapper`, but ended up going another way. Will revert

test/mocks/SimpleERC20Wrapper.sol
```diff
@@ -4,7 +4,6 @@ pragma solidity 0.8.21;
 import {ERC20} from "src/token/ERC20.sol";
 import {IERC20Wrapper, IERC20Metadata} from "src/interfaces/IERC20.sol";
 import "forge-std/Test.sol";
-import {console} from "forge-std/console.sol";
 
 contract SimpleERC20Wrapper is ERC20, IERC20Wrapper {
```
peculiarity, line 8: Do we follow any specific pattern for mocks? Something like adding Mock word as a prefix ? Would be nice to diff somehow the prod code from the one used in tests only.

hieronx: Yeah I didn't use the Mock prefix here because it's not really a mock. It's really a very simple, but technically complete implementation of an ERC20 wrapper.

hieronx: We could also still add Mock prefix to it though, really no strong feelings on this one. Just felt wrong since it's not actually mocking anything.

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,187 @@
...
         require(lockedRequest > 0, "CentrifugeRouter/user-has-no-locked-balance");
-        lockedRequests[msg.sender][vault] = 0;
-        SafeTransferLib.safeTransfer(IERC7540Vault(vault).asset(), msg.sender, lockedRequest);
-        emit UnlockDepositRequest(vault, msg.sender);
+        lockedRequests[_initiator][vault] = 0;
+
+        address asset = IERC7540Vault(vault).asset();
+        EscrowLike(escrow).approve(asset, _initiator, lockedRequest);
+
+        SafeTransferLib.safeTransferFrom(IERC7540Vault(vault).asset(), escrow, _initiator, lockedRequest);
```
peculiarity, line 88:
```suggestion
        SafeTransferLib.safeTransferFrom(asset, escrow, _initiator, lockedRequest);
```

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,187 @@
...
+        require(owner == _initiator || owner == address(this), "CentrifugeRouter/invalid-owner");
+        SafeTransferLib.safeTransferFrom(IERC7540Vault(vault).asset(), owner, escrow, amount);
+        lockedRequests[controller][vault] += amount;
+        emit LockDepositRequest(vault, controller, amount);
     }
 
     /// @inheritdoc ICentrifugeRouter
-    function unlockDepositRequest(address vault) external {
-        uint256 lockedRequest = lockedRequests[msg.sender][vault];
+    function unlockDepositRequest(address vault) external protected {
```
peculiarity, line 80:
```solidity
 function unlockDepositRequest(address vault) external protected {
        _unlock(vault, _initiator);
        emit UnlockDepositRequest(vault, _initiator);
    }

    function executeLockedDepositRequest(address vault, address controller) external protected {
        _unlock(vault, controller);
        _approveMax(asset, vault);
        IERC7540Vault(vault).requestDeposit(lockedRequest, controller, address(this));
        emit ExecuteLockedDepositRequest(vault, controller);
    }
    
function _unlock(address vault, address beneficiary) internal {
        uint256 lockedRequest = lockedRequests[beneficiary][vault];
        require(lockedRequest > 0, "CentrifugeRouter/user-has-no-locked-balance");
        lockedRequests[beneficiary][vault] = 0;

        address asset = IERC7540Vault(vault).asset();
        EscrowLike(escrow).approve(asset, beneficiary, lockedRequest);

        SafeTransferLib.safeTransferFrom(asset, escrow, beneficiary, lockedRequest);
    }    

```

You have to check if I haven't missed anything

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,187 @@
...
+        uint256 lockedRequest = lockedRequests[_initiator][vault];
         require(lockedRequest > 0, "CentrifugeRouter/user-has-no-locked-balance");
-        lockedRequests[msg.sender][vault] = 0;
-        SafeTransferLib.safeTransfer(IERC7540Vault(vault).asset(), msg.sender, lockedRequest);
-        emit UnlockDepositRequest(vault, msg.sender);
+        lockedRequests[_initiator][vault] = 0;
+
+        address asset = IERC7540Vault(vault).asset();
+        EscrowLike(escrow).approve(asset, _initiator, lockedRequest);
```
peculiarity, line 86:
```suggestion
        EscrowLike(escrow).approve(asset, address(this), lockedRequest);
```

peculiarity (dismissed):

peculiarity (approved):

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,197 @@
...
+        IERC7540Vault(vault).requestDeposit(lockedRequest, controller, address(this));
+        emit ExecuteLockedDepositRequest(vault, controller, _initiator);
     }
 
     /// @inheritdoc ICentrifugeRouter
-    function claimDeposit(address vault, address user) external {
-        uint256 maxDeposit = IERC7540Vault(vault).maxDeposit(user);
-        IERC7540Vault(vault).deposit(maxDeposit, user, user);
+    function claimDeposit(address vault, address receiver, address controller) external protected {
+        uint256 maxDeposit = IERC7540Vault(vault).maxDeposit(controller);
```
ilinzweilin, line 119: we have to make sure that the controller is the _initiator, otherwise anybody can pass a random receiver for a controller that has a balance and steal the funds

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,197 @@
...
     /// @inheritdoc ICentrifugeRouter
-    function requestRedeem(address vault, uint256 amount) external {
-        IERC7540Vault(vault).requestRedeem(amount, msg.sender, msg.sender);
+    function requestRedeem(address vault, uint256 amount, address controller, address owner) external protected {
+        IERC7540Vault(vault).requestRedeem(amount, controller, owner);
+    }
+
+    /// @inheritdoc ICentrifugeRouter
+    function claimRedeem(address vault, address receiver, address controller) external protected {
+        uint256 maxRedeem = IERC7540Vault(vault).maxRedeem(controller);
```
ilinzweilin, line 131: same here, we need to make sure that controller is _initiator

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,197 @@
...
-        IERC7540Vault(vault).redeem(maxRedeem, user, user);
+    function permit(address asset, address spender, uint256 assets, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+        external
+        protected
+    {
+        try IERC20Permit(asset).permit(_initiator, spender, assets, deadline, v, r, s) {} catch {}
+    }
+
+    // --- ERC20 wrapping ---
+    function wrap(address wrapper, uint256 amount, address receiver) external protected {
```
ilinzweilin, line 145: let's also include the following check here: 
`require(receiver != address(0), "CentrifugeRouter/zero-address");`
to be consistent with `unwrap`

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,197 @@
...
+        _approveMax(underlying, wrapper);
+        require(IERC20Wrapper(wrapper).depositFor(receiver, amount), "CentrifugeRouter/deposit-for-failed");
+    }
+
+    function unwrap(address wrapper, uint256 amount, address receiver) external protected {
+        require(receiver != address(0), "CentrifugeRouter/zero-address");
+        amount = MathLib.min(amount, IERC20(wrapper).balanceOf(address(this)));
+        require(amount != 0, "CentrifugeRouter/zero-balance");
+
+        require(IERC20Wrapper(wrapper).withdrawTo(receiver, amount), "CentrifugeRouter/withdraw-to-failed");
```
ilinzweilin, line 160: if the transfer of the wraped tokens to the router  happens in a seperate transaction, what prevents a malicious receiver to actually steal the initial receivers balance

ilinzweilin: on `wrap` we actually transfer the tokens to the router, but on `unwrap` we are not transferring the wrapped tokens to the router...

ilinzweilin (changes requested):

src/CentrifugeRouter.sol
peculiarity: So when you use `depositFor`  the `receiver` you passed is actually used in the account balancing in the wrapper. Basically the `balance[receiver]+=amount`

When you try to `withdrawTo` actually it checks the balance of the `msg.sender`.  So the only way to withdraw funds is if the `receiver` in the `wrap` call actually calls `withdrawTo`.

I think there is another issue thought.

When `wrap` is called it will accrue the amount to the `receiver` and the CentrifugeRouter balance will be 0.
When `unwrap` is called it won't pass because the balance of `msg.sender` in the wrapper ( in our case CentrifugeRouter ) must be >= `amount` but it's 0.

I've been checking these 2 implementations ( based on our impl in morpho-markets )
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Wrapper.sol
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol

src/CentrifugeRouter.sol
hieronx: Wow, I'm not sure how I overlooked that, but you're of course right. We can just remove the `controller` praameter here

src/CentrifugeRouter.sol
hieronx: I removed it from unwrap now, it seems like a useless check to me.

src/CentrifugeRouter.sol
hieronx: Fixed in the latest commit. You can now either claim on behalf of someone else, to their address, or on behalf of yourself to any address.

src/CentrifugeRouter.sol
hieronx: I don't fully follow the concern. I added an unwrap test, it works fine from what I see: https://github.com/centrifuge/liquidity-pools/pull/320/commits/83f118b4948fa129602745bffdbf750ea463ec72

src/CentrifugeRouter.sol
```diff
@@ -2,97 +2,197 @@
...
         require(lockedRequest > 0, "CentrifugeRouter/user-has-no-locked-balance");
-        lockedRequests[msg.sender][vault] = 0;
-        SafeTransferLib.safeTransfer(IERC7540Vault(vault).asset(), msg.sender, lockedRequest);
-        emit UnlockDepositRequest(vault, msg.sender);
+        lockedRequests[_initiator][vault] = 0;
+
+        address asset = poolManager.vaultToAsset(vault);
+
+        require(asset != address(0), "CentrifugeRouter/unknown-vault");
+        EscrowLike(escrow).approve(asset, address(this), lockedRequest);
```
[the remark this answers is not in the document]
hieronx: Agreed, updated

peculiarity (approved):

ilinzweilin (approved):

PR #328, opened 2024-06-21
Manage permissionless claiming

hieronx:


AStox (approved):

src/CentrifugeRouter.sol
```diff
@@ -71,6 +74,12 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         emit LockDepositRequest(vault, controller, owner, _initiator, amount);
     }
 
+    /// @inheritdoc ICentrifugeRouter
+    function openLockDepositRequest(address vault, uint256 amount) external protected {
```
hieronx, line 78: Just a helper to allow Prime users that interact with Gnosis Safe directly, to easily construct this transaction.

src/CentrifugeRouter.sol
```diff
@@ -104,7 +113,10 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
 
     /// @inheritdoc ICentrifugeRouter
     function claimDeposit(address vault, address receiver, address controller) external protected {
-        require(controller == _initiator || controller == receiver, "CentrifugeRouter/invalid-sender");
+        require(
+            controller == _initiator || (controller == receiver && opened[controller][vault] == true),
+            "CentrifugeRouter/invalid-sender"
```
peculiarity, line 118: I think you shouldn't  rely on  the controller to call `close` externally but rather  call it here. Otherwise first `claimDeposit` for this controller on this vault is protected but any future operations might be susceptible to the same issue.
Same goes for `claimRedeem`

PR #330, opened 2024-06-23
Refactor tranche token

hieronx:


src/ERC7540Vault.sol
```diff
@@ -136,10 +136,10 @@ contract ERC7540Vault is Auth, IERC7540Vault {
         // the sender is the owner, to bypass the allowance check
         address sender = isOperator[owner][msg.sender] ? owner : msg.sender;
 
-        require(
-            AuthTransferLike(share).authTransferFrom(sender, owner, address(escrow), shares),
-            "ERC7540Vault/transfer-failed"
-        );
+        try AuthTransferLike(share).authTransferFrom(sender, owner, address(escrow), shares) returns (bool success) {}
+        catch {
+            require(IERC20(share).transferFrom(owner, address(escrow), shares), "ERC7540Vault/transfer-from-failed");
```
peculiarity, line 141:
```suggestion            
            require(SafeTransferLib.safeTransferFrom(share, owner, address(escrow), shares), "ERC7540Vault/transfer-from-failed");
```

hieronx: Not required because the share is a trusted token, always our own implementation

src/PoolManager.sol
```diff
@@ -48,6 +52,7 @@ contract PoolManager is Auth, IPoolManager {
 
     IEscrow public immutable escrow;
 
+    address public router;
```
peculiarity, line 55: I don't see this being used anywhere apart of the `file` method. Probably can be deleted?

src/interfaces/token/IRestrictionManager.sol
```diff
@@ -16,11 +16,15 @@ interface IRestrictionManager {
 
     // --- ERC1404 implementation ---
     /// @notice TODO
-    function detectTransferRestriction(address from, address to, uint256 /* value */ ) external view returns (uint8);
+    // function detectTransferRestriction(address from, address to, uint256 /* value */ ) external view returns (uint8);
```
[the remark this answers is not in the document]
hieronx: Yeah the PR is not at this level of refinement ;-) Would be great to get a review of the high-level design changes first

src/interfaces/IERC20.sol
```diff
@@ -175,8 +175,18 @@ interface IERC20Permit {
...
+    uint128 from;
+    uint128 to;
+}
+
 interface IERC20Callback {
-    function onERC20Transfer(address from, address to, uint256 value) external returns (bytes4);
+    function onERC20Transfer(address from, address to, uint256 value, HookData calldata hookdata)
+        external
+        returns (HookData calldata);
+    function onERC20AuthTransfer(address sender, address from, address to, uint256 value, HookData calldata hookdata)
```
peculiarity, line 187: WHy do we return `HookData` ? Is this something we expect to be changed in any of these callback handlers ?

src/PoolManager.sol
```diff
@@ -407,11 +385,9 @@ contract PoolManager is Auth, IPoolManager {
...
             trancheTokenWards
         );
-        address restrictionManager = restrictionManagerFactory.newRestrictionManager(
-            undeployedTranche.restrictionSet, token, restrictionManagerWards
-        );
-
-        TrancheTokenLike(token).file("restrictionManager", restrictionManager);
+        address restrictionManager =
+            restrictionManagerFactory.newRestrictionManager(undeployedTranche.hook, token, restrictionManagerWards);
+        TrancheTokenLike(token).file("hook", restrictionManager);
```
ilinzweilin, line 390: from my understanding, we will add the initial restrictionManager deployment to the deploy script and simply pass to all tranche tokens

src/token/RestrictionManager.sol
```diff
@@ -13,54 +16,100 @@ interface RestrictionManagerLike {
...
         }
 
-        Restrictions memory toRestrictions = restrictions[to];
         bool toIsEndorsed = root.endorsed(to);
-        if (toRestrictions.frozen == true && !toIsEndorsed) {
+        if (hookData.to.getBit(FREEZE_BIT) == true && !toIsEndorsed) {
+            return DESTINATION_IS_FROZEN_CODE;
+        }
+
+        // if (toRestrictions.validUntil < block.timestamp && !toIsEndorsed) {
```
ilinzweilin, line 94: member checks are missing

src/token/RestrictionManager.sol
```diff
@@ -13,54 +16,100 @@ interface RestrictionManagerLike {
...
+        if (hookData.from.getBit(FREEZE_BIT) == true) {
+            return SOURCE_IS_FROZEN_CODE;
+        }
+
+        if (hookData.to.getBit(FREEZE_BIT) == true) {
             return DESTINATION_IS_FROZEN_CODE;
         }
 
-        if (toRestrictions.validUntil < block.timestamp && !toIsEndorsed) {
+        if (hookData.to.getBit(MEMBER_BIT) == false) {
```
ilinzweilin, line 112: check for valid member missing

ilinzweilin (approved):
looks good. Member checks still missing though

src/token/Tranche.sol
```diff
@@ -13,32 +14,50 @@ interface TrancheTokenLike is IERC20Metadata {
...
+    modifier authOrHook() {
+        require(wards[msg.sender] == 1 || msg.sender == hook, "Auth/not-authorized");
+        _;
+    }
 
     // --- Administration ---
     /// @inheritdoc ITrancheToken
-    function file(bytes32 what, address data) external auth {
-        if (what == "restrictionManager") restrictionManager = data;
+    function file(bytes32 what, address data) external authOrHook {
```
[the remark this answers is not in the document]
hieronx: This allows a hook to implement migration logic, being able to migrate from 1 hook to another.

(will of course add this as a comment)

src/PoolManager.sol
hieronx: What was  rather my idea is that we add an `address` arg to the `AddTranche` message, replacing the `restrictionSet`. So we always pass the restriction manager address here

src/token/RestrictionManager.sol
hieronx: Fixed!

src/interfaces/IERC20.sol
hieronx: That was the idea, but I realized it does not make sense. Removed!

.github/workflows/ci.yml
```diff
@@ -24,19 +24,19 @@ jobs:
           FOUNDRY_PROFILE: ci
           FORK_TESTS: false
 
-  test-invariant:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
+  # test-invariant:
```
hieronx, line 27: It doesn't seem worth fixing these as they anyways will be replaced by the new Recon based invariant tests soon.

src/PoolManager.sol
```diff
@@ -476,9 +425,10 @@ contract PoolManager is Auth, IPoolManager {
         emit RemoveVault(poolId, trancheId, asset, vault);
     }
 
-    function updateCentrifugeGasPrice(uint256 price, uint256 computedAt) public auth {
+    /// @inheritdoc IPoolManager
+    function updateCentrifugeGasPrice(uint128 price, uint256 computedAt) public auth {
         require(price > 0, "PoolManager/price-cannot-be-zero");
-        require(gasService.price() != price, "PoolManager/same-price-already-set");
+        require(gasService.gasPrice() != price, "PoolManager/same-price-already-set");
```
hieronx, line 431: @peculiarity found a little bug here while importing the correct interface

src/Root.sol
```diff
@@ -124,7 +124,7 @@ contract Root is Auth, IRoot {
         } else if (call == MessagesLib.Call.RecoverTokens) {
             (address target, address token, address to, uint256 amount) =
                 (message.toAddress(1), message.toAddress(33), message.toAddress(65), message.toUint256(97));
-            RecoverLike(target).recoverTokens(token, to, amount);
+            recoverTokens(target, token, to, amount);
```
hieronx, line 127: Unrelated to the PR but this was off (no reason to have the internal method otherwise)

src/libraries/BitmapLib.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.21;
+
+/// @title  BitmapLib
+library BitmapLib {
```
hieronx, line 5: Added a ticket for writing unit tests just for the library: https://github.com/centrifuge/liquidity-pools/issues/339

test/unit/token/ERC20.t.sol
```diff
@@ -496,26 +496,6 @@ contract ERC20Test is Test {
         }
     }
 
-    function testAuthTransferFrom(address to, uint256 amount) public {
```
hieronx, line 499: Moved to `Tranche.t.sol`

test/unit/token/Tranche.t.sol
```diff
@@ -175,47 +175,24 @@ contract TrancheTokenTest is Test {
...
         assertEq(token.balanceOf(targetUser), amount);
         afterTransferAssumptions(address(0), targetUser, amount);
 
         vm.warp(validUntil + 1);
 
-        vm.expectRevert(bytes("RestrictionManager/destination-not-a-member"));
-        token.mint(targetUser, amount);
-    }
-
-    function testTransferMintFailsNoPermissionOnRestrictionManager() public {
```
hieronx, line 200: Does not apply anymore since the tranche token is not anymore a ward on the hook/restriction manager

script/Deployer.sol
```diff
@@ -52,29 +52,29 @@ contract Deployer is Script {
         root = new Root{salt: salt}(address(escrow), delay, deployer);
 
         vaultFactory = address(new ERC7540VaultFactory(address(root)));
-        restrictionManagerFactory = address(new RestrictionManagerFactory(address(root)));
+        restrictionManager = address(new RestrictionManager{salt: salt}(address(root)));
```
hieronx, line 55: Deployed deterministically so Centrifuge Chain can store 1 restriction manager address as the default for all chains

src/PoolManager.sol
```diff
@@ -314,34 +300,10 @@ contract PoolManager is Auth, IPoolManager {
...
-
-        RestrictionManagerLike restrictionManager = RestrictionManagerLike(address(trancheToken.restrictionManager()));
-        restrictionManager.updateMember(user, validUntil);
-    }
-
-    /// @inheritdoc IPoolManager
-    function freeze(uint64 poolId, bytes16 trancheId, address user) public auth {
-        require(user != address(escrow), "PoolManager/escrow-cannot-be-frozen");
-
+    function updateRestriction(uint64 poolId, bytes16 trancheId, bytes memory update) public auth {
```
hieronx, line 303: Changed to an `UpdateRestriction` message to be more generic, and more easily allow adding custom restriction types for future hooks.

src/InvestmentManager.sol
```diff
@@ -11,31 +11,9 @@ import {IRoot} from "src/interfaces/IRoot.sol";
 import {IERC20, IERC20Metadata} from "src/interfaces/IERC20.sol";
 import {IPoolManager} from "src/interfaces/IPoolManager.sol";
 import {IInvestmentManager, InvestmentState} from "src/interfaces/IInvestmentManager.sol";
-
-interface GatewayLike {
```
hieronx, line 15: Here, and in a lot of other files, I removed the duplicate interfaces and just used the existing ones.

src/PoolManager.sol
```diff
@@ -360,7 +317,11 @@ contract PoolManager is Auth, IPoolManager {
 
         // Give investment manager infinite approval for asset
         // in the escrow to transfer to the user on redeem or withdraw
-        escrow.approveMax(asset, address(investmentManager));
+        escrow.approveMax(asset, investmentManager);
+
+        // Give pool manager infinite approval for asset
+        // in the escrow to transfer to the user on transfer
+        escrow.approveMax(asset, address(this));
```
hieronx, line 324: This was in `handleTransfer` every time, it's more gas efficient to just do it once.

src/PoolManager.sol
```diff
@@ -379,12 +339,19 @@ contract PoolManager is Auth, IPoolManager {
...
         auth
     {
-        TrancheTokenLike trancheToken = TrancheTokenLike(getTrancheToken(poolId, trancheId));
+        ITrancheToken trancheToken = ITrancheToken(getTrancheToken(poolId, trancheId));
         require(address(trancheToken) != address(0), "PoolManager/unknown-token");
 
         trancheToken.mint(destinationAddress, amount);
     }
 
+    /// @inheritdoc IPoolManager
```
hieronx, line 348: Just moved it to the right section

src/PoolManager.sol
```diff
@@ -432,26 +392,22 @@ contract PoolManager is Auth, IPoolManager {
...
         );
         vaultToAsset[vault] = asset;
 
         // Link vault to tranche token
         AuthLike(tranche.token).rely(vault);
-        TrancheTokenLike(tranche.token).updateVault(asset, vault);
-
-        // Give vault infinite approval for tranche tokens
-        // in the escrow to burn on executed redemptions
-        escrow.approveMax(tranche.token, vault);
```
hieronx, line 454: This was actually not necessary

src/token/Tranche.sol
```diff
@@ -2,100 +2,149 @@
...
-    address public restrictionManager;
+/// @notice Extension of ERC20 + ERC1404 for tranche tokens,
+///         integrating an external hook optionally for ERC20 callbacks and ERC1404 checks.
+///
+/// @dev    The user balance is limited to uint128. This is safe because the decimals are limited to 18,
+///         thus the max balance is 2^128-1 / 10**18 = 3.40 * 10**20. This is also enforced on mint.
+///
+///         The most significant 128 bits of the uint256 balance value are used
+///         to store hook data (e.g. restrictions for users).
+contract TrancheToken is ERC20, ITrancheToken {
```
peculiarity, line 20: nit: I have the feeling that we can simply use `Tranche` and `ITranche`.

src/CentrifugeRouter.sol
```diff
@@ -9,27 +9,23 @@ import {IERC7540Vault} from "src/interfaces/IERC7540.sol";
...
     address constant UNSET_INITIATOR = address(1);
     address internal _initiator = UNSET_INITIATOR;
 
     /// @inheritdoc ICentrifugeRouter
     mapping(address controller => mapping(address vault => uint256 amount)) public lockedRequests;
 
-    constructor(address escrow_, address poolManager_, address gateway_) {
+    constructor(address escrow_, address gateway_, address poolManager_) {
         escrow = IEscrow(escrow_);
+        gateway = IGateway(gateway_);
```
peculiarity, line 27: Any particular reason why we switched from "in-contract" `GatewayLike` to full-blown `IGateway` ?

Using the `GatewayLike` vs `import IGateway` approach
Pros:
 - Not all code that is defined in the interface gets copied into the file that imports the interface. - Grows deployment size.
 - Expose only functions that are meant to be called within the scope of the contract that's using it.  - Might result in calling something that is unexpected.
 
 Cons: 
 - Every single time you change a signature, you have to go through all `GatewayLike` definitions and change the signature there as well which is time consuming and annoying.

src/interfaces/token/IHook.sol
```diff
@@ -0,0 +1,36 @@
...
+
+struct HookData {
+    bytes16 from;
+    bytes16 to;
+}
+
+uint8 constant SUCCESS_CODE = 0;
+string constant SUCCESS_MESSAGE = "transfer-allowed";
+
+interface IHook is IERC165 {
```
peculiarity, line 14: I was thinking that `Hook` is a very generic term. Here we define an interface of a hook that has very specific  functions.
What do you think about defining it like [that](https://docs.stackup.sh/docs/erc-7579#hooks) 

Then we can have  multiple multiple implementations such as `AuthTransferHook` , `TransferHook` , `RestrictionsHook` ( which is the manager itself  ). 

Also in the `TrancheToken` you won't have a single `hook`  but rather a mapping of hooks  mapped by name and address and you can call different hook within the `TrancheToken` based on their name and just call `preCheck` , `postCheck`.

This is a very rough idea.

src/token/RestrictionManager.sol
```diff
@@ -2,72 +2,94 @@
...
+///         * Supports freezing accounts which blocks transfers both to and from them
+///         * Allows authTransferFrom calls
+///
+/// @dev    The first 8 bytes (uint64) of hookData is used for the memberlist valid until date,
+///         the last bit is used to denote whether the account is frozen.
+contract RestrictionManager is Auth, IRestrictionManager, IHook {
+    using BitmapLib for *;
+    using BytesLib for bytes;
+
+    string internal constant SOURCE_IS_FROZEN_MESSAGE = "source-is-frozen";
```
peculiarity, line 27: Aren't these supposed to be imported from `IRestrictionManager` ?

src/libraries/BitmapLib.sol
```diff
@@ -0,0 +1,43 @@
...
+    function setBit(uint128 bitmap, uint128 index, bool isTrue) internal pure returns (uint128) {
+        if (isTrue) {
+            return bitmap | (uint128(1) << index);
+        } else {
+            return bitmap & ~(uint128(1) << index);
+        }
+    }
+
+    function getBit(uint128 bitmap, uint128 index) internal pure returns (bool) {
```
peculiarity, line 14: I don't like this. `getBit` returns `bool`. It's a little bit fishy. If you read somewhere within the code `getBit` you expect a number to be returned. Instead you see `bool` and start wondering ... How does this work? Why does it return bool? 
My suggestion would be to either rename it to `isOne` or return a number :)

src/token/RestrictionManager.sol
```diff
@@ -84,33 +106,64 @@ contract RestrictionManager is Auth, IRestrictionManager, IERC20Callback {
...
     /// @inheritdoc IRestrictionManager
-    function updateMember(address user, uint64 validUntil) public auth {
+    function updateMember(address token, address user, uint64 validUntil) public auth {
         require(block.timestamp <= validUntil, "RestrictionManager/invalid-valid-until");
-        restrictions[user].validUntil = validUntil;
+        require(!root.endorsed(user), "RestrictionManager/endorsed-user-cannot-be-updated");
+        _updateMember(token, user, validUntil);
+    }
+
+    function _updateMember(address token, address user, uint64 validUntil) internal {
```
peculiarity, line 152: What's the benefit of this internal method? Why not inline the body of this function in `updateMember`?

test/unit/token/Tranche.t.sol
```diff
@@ -175,47 +175,24 @@ contract TrancheTokenTest is Test {
         assertEq(token.balanceOf(self), amount);
     }
 
-    // mint
+    // // mint
```
peculiarity, line 178: I think you need to add `//` one more time  :) /s

src/token/RestrictionManager.sol
hieronx: Good catch, this was required in a previous draft but not anymore, will change

src/PoolManager.sol
```diff
@@ -407,19 +368,15 @@ contract PoolManager is Auth, IPoolManager {
             undeployedTranche.decimals,
             trancheTokenWards
         );
-        address restrictionManager = restrictionManagerFactory.newRestrictionManager(
-            undeployedTranche.restrictionSet, token, restrictionManagerWards
-        );
-
-        TrancheTokenLike(token).file("restrictionManager", restrictionManager);
+        ITrancheToken(token).file("hook", undeployedTranche.hook);
```
ilinzweilin, line 371: I think we need to require that the hook address is set

ilinzweilin (dismissed):

src/PoolManager.sol
hieronx: What do you mean?

ilinzweilin: when you add the tranche, the hook can be set to address(0)
https://github.com/centrifuge/liquidity-pools/blob/c6dce19190bb7dc32372830a0be410b4ba6c7d3c/src/PoolManager.sol#L245

Should we allow deployments without a valid hook address?

hieronx: The comment in line 244 answers that :)

peculiarity (approved):

ilinzweilin (approved):

PR #331, opened 2024-06-24
chore: Move responsibility fo handling an exception from the caller t…

peculiarity:
…o the callee ( method itself )

### DIFF
--- a/src/interfaces/IPoolManager.sol
+++ b/src/interfaces/IPoolManager.sol
@@ -185,6 +185,13 @@ interface IPoolManager {
         view
         returns (uint128 price, uint64 computedAt);
 
+    /// @notice Function to get the vault's underlying asset
+    /// @dev function vaultToAsset which is a state variable getter could be used
+    /// but in that case each caller MUST make sure they handle the case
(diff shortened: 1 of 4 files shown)


src/interfaces/IPoolManager.sol
```diff
@@ -185,6 +185,13 @@ interface IPoolManager {
         view
         returns (uint128 price, uint64 computedAt);
 
+    /// @notice Function to get the vault's underlying asset
+    /// @dev function vaultToAsset which is a state variable getter could be used
+    /// but in that case each caller MUST make sure they handle the case
+    /// where a 0 address is returned. Using this method, that handling is done
+    /// on the behalf the caller.
```
hieronx, line 192:
```suggestion
    /// @notice Function to get the vault's underlying asset
    /// @dev     Function vaultToAsset which is a state variable getter could be used
    ///                but in that case each caller MUST make sure they handle the case
    ///                where a 0 address is returned. Using this method, that handling is done
    ///                on the behalf the caller.
```

hieronx (approved):

PR #335, opened 2024-06-25
fea:(Gateway) Allow partial payments from the gateway balance for transaction not triggered by CentrifugeRouter

peculiarity:
1. Remove `test/unit/GatewayExtended.t.sol`  and move almost all tests into `test/unit/Gateway.t.sol`
 
 2. Create `test/integration/Gateway.t.sol` and test the deployment only there
 
 3.  Write a complete* set of unit tests for gas payments for the `Gateway.send`.
          - `testOutgoingMessagingWithNotEnoughPrepayment` - ( `fuel < estimate` ) 
          - `testOutgoingMessagingWithPrepayment` - ( `fuel = estimate` )
          - `testOutgoingMessagingWithExtraPrepayment` - ( `fuel > estimate` )
          - `testingOutgoingMessagingWithCoveredPayment` - `( fuel = 0 , gateway.balance >= estimate )`
          - `testingOutgoingMessagingWithPartiallyCoveredPayment` - `( fuel = 0, gateway.balance > 0, gateway.balance < estimate )`
          - `testingOutgoingMessagingWithoutBeingCovered` - `( fuel = 0 , gateway.balance = 0) `
          - `testingOutgoingMessagingWherePaymentCoverIsNotAllowed` - `(fuel = 0 , gateway.balance - irrelevant, shouldRefuel = false)`
 
4. Write also extensive unit for `topUp` function - `testPrepayment`.
 
 * - at least to me knowledge :D


src/gateway/Gateway.sol
```diff
@@ -254,26 +254,47 @@ contract Gateway is Auth, IGateway {
...
+
+                currentRouter.pay{value: consumed}(payload, address(this));
 
-        if (fuel > 0) quota = 0;
+                currentRouter.send(payload);
+            }
+            quota = 0;
+        } else if (gasService.shouldRefuel(source, message)) {
+            uint256 tank = address(this).balance;
+            for (uint256 i; i < numRouters; i++) {
```
peculiarity, line 281: I know this loop is duplicated  but IMO the way it's structured like this makes it more readable and shows clearer intention. I'm happy to discuss it though. 
The body of the loop is not completely the same but the loop itself can be extracted in a separate function and just  pass some payment handling logic. I think it will get more complicated

hieronx: Yeah I went over this 3 times and came the same conclusion so far, it's not ideal but doesn't seem like any alternatives are cleaner :)

hieronx (approved):

hieronx (approved):

PR #342, opened 2024-07-02
Fix restriction manager

hieronx:


test/unit/token/RestrictionManager.t.sol
```diff
@@ -48,6 +52,22 @@ contract RestrictionManagerTest is Test {
...
+        restrictionManager.updateMember(address(token), address(this), validUntil);
+        (bool _isMember, uint64 _validUntil) = restrictionManager.isMember(address(token), address(this));
+        assertTrue(_isMember);
+        assertEq(_validUntil, validUntil);
+        assertEq(restrictionManager.isFrozen(address(token), address(this)), false);
+
+        restrictionManager.freeze(address(token), address(this));
+        (_isMember, _validUntil) = restrictionManager.isMember(address(token), address(this));
+        assertTrue(_isMember);
+        assertEq(_validUntil, validUntil);
```
hieronx, line 67: This caught the freeze bit being incorrect, since valid until would be overwritten.

PR #343, opened 2024-07-03
Add vault functions to router

AStox:
https://github.com/centrifuge/liquidity-pools/issues/314
https://github.com/centrifuge/liquidity-pools/issues/322


src/CentrifugeRouter.sol
```diff
@@ -8,9 +8,19 @@ import {IERC20, IERC20Permit, IERC20Wrapper} from "src/interfaces/IERC20.sol";
 import {IERC7540Vault} from "src/interfaces/IERC7540.sol";
 import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
 import {IPoolManager} from "src/interfaces/IPoolManager.sol";
+import {IInvestmentManager} from "src/interfaces/IInvestmentManager.sol";
 import {IEscrow} from "src/interfaces/IEscrow.sol";
 import {IGateway} from "src/interfaces/gateway/IGateway.sol";
 
+interface GatewayLike {
```
peculiarity, line 15: This shouldn't be here.  As you can see `IGateway` is imported
`

src/CentrifugeRouter.sol
```diff
@@ -140,6 +150,18 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         IERC7540Vault(vault).deposit(maxDeposit, receiver, controller);
     }
 
+    /// @inheritdoc ICentrifugeRouter
+    function cancelDepositRequest(address vault, address controller) external payable protected {
+        validateController(vault, controller);
```
peculiarity, line 155: Do we need `validateController` check in the route at all ? It's a duplicate and it's already being done at ERC7540 level :S

src/CentrifugeRouter.sol
```diff
@@ -208,6 +276,15 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         require(IERC20Wrapper(wrapper).withdrawTo(receiver, amount), "CentrifugeRouter/withdraw-to-failed");
     }
 
+    // --- ERC20 auth transfer ---
+    /// @inheritdoc ICentrifugeRouter
+    function authTransferFrom(address vault, address sender, address owner, address recipient, uint256 amount)
```
peculiarity, line 281: Not sure if we need this here either.

src/CentrifugeRouter.sol
```diff
@@ -169,6 +191,52 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
+
+    /// @inheritdoc ICentrifugeRouter
+    function claimCancelRedeemRequest(address vault, address receiver, address controller) external payable protected {
+        validateController(vault, controller);
+        IERC7540Vault(vault).claimCancelRedeemRequest(0, receiver, controller);
+    }
+
+    // --- Transfer ---
+    /// @inheritdoc ICentrifugeRouter
+    function transfer(address asset, bytes32 recipient, uint128 amount) external protected {
```
peculiarity, line 208: it will be user friendlier if it accepts `address` and just `toBytes32()` when you call `poolManager.transfer`

src/CentrifugeRouter.sol
```diff
@@ -169,6 +191,52 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
+    }
+
+    /// @inheritdoc ICentrifugeRouter
+    function transferTranchesToEVM(address vault, uint64 destinationChainId, address destinationAddress, uint128 amount)
+        external
+        payable
+        protected
+    {
+        SafeTransferLib.safeTransferFrom(IERC7540Vault(vault).share(), msg.sender, address(this), amount);
+        IERC20(IERC7540Vault(vault).share()).approve(address(poolManager), amount);
```
peculiarity, line 221: I think we can have `_approveMax` called here

src/CentrifugeRouter.sol
```diff
@@ -169,6 +191,52 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
+            destinationChainId,
+            destinationAddress,
+            amount
+        );
+    }
+
+    /// @inheritdoc ICentrifugeRouter
+    function transferTranchesToCentrifuge(address vault, bytes32 destinationAddress, uint128 amount) external payable protected {
+        SafeTransferLib.safeTransferFrom(IERC7540Vault(vault).share(), msg.sender, address(this), amount);
+        IERC20(IERC7540Vault(vault).share()).approve(address(poolManager), amount);
```
peculiarity, line 234: `_approveMax`

src/CentrifugeRouter.sol
AStox: I'm also confused about this tbh. It's mentioned in the ticket directly but I dont understand why. @hieronx could you shed some light?

src/CentrifugeRouter.sol
AStox: `validateController` gets called at the vault level, but then it's checking whether `router` is a valid controller. It's endorsed so it always passes this check, so without also validating at the `router` level anybody could call this on anyone else's behalf

src/CentrifugeRouter.sol
hieronx: Yeah my bad, `authTransferFrom` here makes no sense

src/CentrifugeRouter.sol
hieronx: This doesn't work. For `cancelDeposit/RedeemRequest`, it doesn't make sense to pass a `controller`, it should just always use `_initiatior()`

hieronx: For `claimCancelDeposit/RedeemRequest`, you should copy the logic from `claimDeposit/RedeemRequest`:
```solidity
require(
            controller == _initiator || (controller == receiver && opened[controller][vault] == true),
            "CentrifugeRouter/invalid-sender"
        );
```

src/CentrifugeRouter.sol
```diff
@@ -169,6 +186,52 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
+
+    /// @inheritdoc ICentrifugeRouter
+    function claimCancelRedeemRequest(address vault, address receiver, address controller) external payable protected {
+        validateController(vault, controller);
+        IERC7540Vault(vault).claimCancelRedeemRequest(0, receiver, controller);
+    }
+
+    // --- Transfer ---
+    /// @inheritdoc ICentrifugeRouter
+    function transfer(address asset, address recipient, uint128 amount) external protected {
```
hieronx, line 203: I think here and also in pool mgr, we should actually rename this to `transferAssets`, to be more explicit

src/CentrifugeRouter.sol
hieronx: @peculiarity @AStox this doesn't work because it's transferring to Centrifuge Chain, where addresses are 32 bytes

src/CentrifugeRouter.sol
hieronx: https://github.com/centrifuge/liquidity-pools/pull/355 put up a bit of a transfer refactor here

src/CentrifugeRouter.sol
```diff
@@ -4,14 +4,22 @@ pragma solidity 0.8.26;
...
-import {IPoolManager} from "src/interfaces/IPoolManager.sol";
+import {IPoolManager, Domain} from "src/interfaces/IPoolManager.sol";
+import {IInvestmentManager} from "src/interfaces/IInvestmentManager.sol";
 import {IEscrow} from "src/interfaces/IEscrow.sol";
 import {IGateway} from "src/interfaces/gateway/IGateway.sol";
 
+interface AuthTransferLike {
+    function authTransferFrom(address sender, address owner, address recipient, uint256 amount) external;
+}
+
```
hieronx, line 19:
```suggestion
```

src/CentrifugeRouter.sol
hieronx: > For cancelDeposit/RedeemRequest, it doesn't make sense to pass a controller, it should just always use _initiatior()

@AStox see this line above, that is not addressed correctly now

src/CentrifugeRouter.sol
```diff
@@ -169,6 +207,79 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         }
     }
 
+    /// @inheritdoc ICentrifugeRouter
+    function cancelRedeemRequest(address vault, address controller, uint256 topUpAmount) external payable protected {
+        require(topUpAmount <= address(this).balance, "CentrifugeRouter/insufficient-funds-to-topup");
+
+        validateController(vault, controller);
```
hieronx, line 214: Likewise as `cancelDepositRequest`

src/ERC7540Vault.sol
```diff
@@ -404,4 +404,8 @@ contract ERC7540Vault is Auth, IERC7540Vault {
             "ERC7540Vault/invalid-controller"
         );
     }
+
+    function authTransferFrom(address sender, address owner, address recipient, uint256 shares) external auth {
+        ITranche(share).authTransferFrom(sender, owner, recipient, shares);
+    }
```
hieronx, line 410:
```suggestion
```

src/interfaces/IERC7540.sol
```diff
@@ -295,4 +295,7 @@ interface IERC7540Vault is
 
     /// @notice TODO
     function onCancelRedeemClaimable(address owner, uint256 shares) external;
+
+    /// @notice TODO
+    function authTransferFrom(address sender, address owner, address recipient, uint256 shares) external;
```
hieronx, line 300:
```suggestion
```

src/CentrifugeRouter.sol
hieronx: Yeah exactly

hieronx (approved):

peculiarity (approved):

PR #345, opened 2024-07-03
Use transient storage in router and gateway

hieronx:


peculiarity (approved):

peculiarity (dismissed):

peculiarity (dismissed):

src/CentrifugeRouter.sol
AStox, line 1: Neat!

src/gateway/Gateway.sol
AStox, line 1: Can't comment directly on the line, but line 45 initializes an unused `uint256 quota` which I think is unused

hieronx: Good catch!

src/CentrifugeRouter.sol
```diff
@@ -10,15 +10,18 @@ import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
 import {IPoolManager} from "src/interfaces/IPoolManager.sol";
 import {IEscrow} from "src/interfaces/IEscrow.sol";
 import {IGateway} from "src/interfaces/gateway/IGateway.sol";
+import {TransientStorage} from "src/libraries/TransientStorage.sol";
 
 contract CentrifugeRouter is Auth, ICentrifugeRouter {
+    using TransientStorage for bytes32;
+
+    // The slot holding the initiator state, transiently. bytes32(uint256(keccak256("initiator")) - 1)
+    bytes32 public constant INITIATOR_SLOT = 0x390f14ca25a428cfdaf9fa3f25a94d06d247d3fa36c5a914ca10e13a4120db3c;
```
peculiarity, line 19: why having a comment and hardcoded value when you can have 
` bytes32 public constant INITIATOR_SLOT = bytes32(uint256(keccak256("initiator")) - 1)`
Constants are evaluated during compile time.

AStox (approved):

src/libraries/TransientStorage.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.26;
+
+/// @title  TransientStorage
+library TransientStorage {
+    function tstore(bytes32 slot, address value) internal {
```
peculiarity, line 6: why do we have to name `tstore` and `tload`?

src/libraries/TransientStorage.sol
```diff
@@ -0,0 +1,29 @@
...
+        }
+    }
+
+    function tstore(bytes32 slot, uint256 value) internal {
+        assembly {
+            tstore(slot, value)
+        }
+    }
+
+    function tloadAddress(bytes32 slot) internal returns (address value) {
```
peculiarity, line 18:
```suggestion
    function asAddress(bytes32 slot) internal returns (address value) {
```

src/libraries/TransientStorage.sol
```diff
@@ -0,0 +1,29 @@
...
+        assembly {
+            value := tload(slot)
+        }
+    }
+
+    function tloadUint256(bytes32 slot) internal returns (uint256 value) {
```
peculiarity, line 24: asUint256

src/CentrifugeRouter.sol
```diff
@@ -32,14 +35,15 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
     modifier protected() {
-        if (_initiator == UNSET_INITIATOR) {
+        address currentInitiator = _initiator();
+        if (currentInitiator == address(0)) {
             // Single call re-entrancy lock
-            _initiator = msg.sender;
+            INITIATOR_SLOT.tstore(msg.sender);
             _;
-            _initiator = UNSET_INITIATOR;
+            INITIATOR_SLOT.tstore(address(0));
```
peculiarity, line 43: Not sure if we need this line which is 100 gas more.
After each transaction the values in tstorage are discarded. The wyay I understand it is that the value are back to default ones

src/CentrifugeRouter.sol
```diff
@@ -83,13 +87,13 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         payable
         protected
     {
```
peculiarity, line 89:
```suggestion
    {
      address initiator = _initiator()
```

src/CentrifugeRouter.sol
```diff
@@ -83,13 +87,13 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
-        require(owner == _initiator || owner == address(this), "CentrifugeRouter/invalid-owner");
+        require(owner == _initiator() || owner == address(this), "CentrifugeRouter/invalid-owner");
```
peculiarity, line 90: use the memory ref.

src/CentrifugeRouter.sol
```diff
@@ -99,24 +103,24 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
         (address asset, bool isWrapper) = poolManager.getVaultAsset(vault);
```
peculiarity, line 104:
```suggestion
address initiator = initiator()
```

src/CentrifugeRouter.sol
```diff
@@ -99,24 +103,24 @@ contract CentrifugeRouter is Auth, ICentrifugeRouter {
...
+            lockDepositRequest(vault, amount, _initiator(), address(this));
         } else {
-            lockDepositRequest(vault, amount, _initiator, _initiator);
+            lockDepositRequest(vault, amount, _initiator(), _initiator());
         }
     }
 
     /// @inheritdoc ICentrifugeRouter
     function unlockDepositRequest(address vault, address receiver) external payable protected {
-        uint256 lockedRequest = lockedRequests[_initiator][vault];
```
peculiarity, line 111: Again - memory ref `address initiator = `initiator()`

src/CentrifugeRouter.sol
```diff
@@ -10,15 +10,18 @@ import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
 import {IPoolManager} from "src/interfaces/IPoolManager.sol";
 import {IEscrow} from "src/interfaces/IEscrow.sol";
 import {IGateway} from "src/interfaces/gateway/IGateway.sol";
+import {TransientStorage} from "src/libraries/TransientStorage.sol";
```
peculiarity, line 14:
```suggestion
type Slot is bytes32;
```

src/CentrifugeRouter.sol
```diff
@@ -10,15 +10,18 @@ import {ICentrifugeRouter} from "src/interfaces/ICentrifugeRouter.sol";
...
 
 contract CentrifugeRouter is Auth, ICentrifugeRouter {
+    using TransientStorage for bytes32;
```
peculiarity, line 16:
```suggestion
    using TransientStorage for Slot;
```

src/CentrifugeRouter.sol
peculiarity, line 19:
```suggestion
    Slot public constant INITIATOR_SLOT = 0x390f14ca25a428cfdaf9fa3f25a94d06d247d3fa36c5a914ca10e13a4120db3c;
```

src/libraries/TransientStorage.sol
hieronx: I think it is valuable to name this explicitly. Also what Uniswap/Balancer/OZ are all doing.

src/CentrifugeRouter.sol
hieronx: True, should not be needed now that we are using transient storage. Good find

src/libraries/TransientStorage.sol
hieronx: Likewise I think this is good to name explicitly

src/CentrifugeRouter.sol
hieronx: @peculiarity a lot of tests are failing if I remove this. I guess the test case is 1 transaction, and if you do multiple calls separately to the router within 1 test, you need the reset above.

I guess this is also the reason you need it anyways: if some contract wants to call the router several times in 1 method (although it does not make much sense because you have multicall), you need the same.

So I would keep it then

src/CentrifugeRouter.sol
hieronx: True, fixed!

src/CentrifugeRouter.sol
peculiarity: Yeah, I wasn't really sure. Let's leave it then

ilinzweilin (approved):

src/gateway/Gateway.sol
```diff
@@ -20,6 +21,10 @@ import {IAdapter} from "src/interfaces/gateway/IAdapter.sol";
 contract Gateway is Auth, IGateway {
     using ArrayLib for uint16[8];
     using BytesLib for bytes;
+    using TransientStorage for bytes32;
+
+    // The slot holding the quota state, transiently. bytes32(uint256(keccak256("quota")) - 1)
+    bytes32 public constant QUOTA_SLOT = 0x1b6c99859b82987bd128ac509391b5af30e732c101d06c7836845a4a5b8e14f6;
```
peculiarity, line 27:
```suggestion
    bytes32 public constant QUOTA_SLOT = bytes32(uint256(keccak256("quota")) - 1);
```

peculiarity (approved):

PR #364, opened 2024-07-10
Guardian coverage to 💯

AStox:
This PR...
1. updates a bug found in `Guardian.sol` where `disputeMessageRecovery` was using an outdated interface.
2. adds a `testDisputeMessageRecovery` integration test
3. adds a `testGuardian` unit test that just tests the constructor in order to bump the coverage up from 72% to 100%


src/admin/Guardian.sol
```diff
@@ -57,8 +54,8 @@ contract Guardian is IGuardian {
     }
 
     /// @inheritdoc IGuardian
-    function disputeMessageRecovery(bytes32 messageHash) external onlySafe {
-        gateway.disputeMessageRecovery(messageHash);
+    function disputeMessageRecovery(address adapter, bytes32 messageHash) external onlySafe {
```
hieronx, line 57: Looks like the only thing that is not tested yet, is that this method can only be called by `safe`. Do you want to add that as well?

AStox: I instead added unit tests to generally test the `onlySafe` and `onlySafeAndOwner` modifiers

hieronx (dismissed):

test/integration/Admin.t.sol
```diff
@@ -2,10 +2,30 @@
...
+import {CastLib} from "src/libraries/CastLib.sol";
 
 contract AdminTest is BaseTest {
+    using CastLib for *;
+
+    address[] threeMockAdapters;
+
     function setUp() public override {
         super.setUp();
+        adapter1 = new MockAdapter(address(gateway));
```
peculiarity, line 15: Since you are using the `super.setUp()` and you inherit from the `BaseTest` you can use the `testAdapters`

peculiarity (approved):

hieronx (approved):

PR #381, opened 2024-07-17
Fix local adapter

hieronx:


wischli (approved):
Thanks so much for this blazing fast fix! It works like a charm for our integration tests!

PR #383, opened 2024-07-22
(#34) Verify vault addresses in Investment Manager

AStox:
[#34](https://cantina.xyz/code/bf6d8467-50b0-49b2-892b-267ea8185b3a/findings/34)

### DIFF
--- a/src/InvestmentManager.sol
+++ b/src/InvestmentManager.sol
@@ -245,6 +245,7 @@ contract InvestmentManager is Auth, IInvestmentManager {
         uint128 shares
     ) public auth {
         address vault = poolManager.getVault(poolId, trancheId, assetId);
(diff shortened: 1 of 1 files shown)


src/InvestmentManager.sol
```diff
@@ -245,6 +245,7 @@ contract InvestmentManager is Auth, IInvestmentManager {
         uint128 shares
     ) public auth {
         address vault = poolManager.getVault(poolId, trancheId, assetId);
+        require(vault != address(0), "InvestmentManager/vault-not-found");
```
hieronx, line 248: I think we can add this check to `PoolManager.getVault`, similar to what we do in `PoolManager.getVaultAsset`

PR #397, opened 2024-08-12
Setup router invariants

hieronx:


test/recon-core/CryticToFoundry.sol
```diff
@@ -36,4 +36,14 @@ contract CryticToFoundry is Test, TargetFunctions, FoundryAsserts {
         poolManager_freeze();
         erc7540_6_deposit_call_target(72);
     }
+
+    // forge test --match-test test_invariant_E_1_0 -vv
+    function test_invariant_E_1_0() public {
+        deployNewTokenPoolAndTranche(81, 113078215817378582107422693752520426616851634649913195634947047953023656943);
```
peculiarity, line 42: Would be nice to assign some variable ( meaningful name ) to these magic numbers :D.
Wdyt?

hieronx: These are tests that are copy pasted as output of broken properties from medusa/echidna. So there is no meaningful names for these, and we shouldn't adapt these.

PR #398, opened 2024-08-12
Add more router tests

AStox:
This PR adds another multicall test to reflect how the frontend will be using multicalls.
It also adds a test to verify users can call unlockDepositRequest while the asset is disallowed on the pool.

### DIFF
--- a/test/integration/CentrifugeRouter.t.sol
+++ b/test/integration/CentrifugeRouter.t.sol
@@ -270,6 +270,34 @@ contract CentrifugeRouterTest is BaseTest {
         assertApproxEqAbs(erc20Y.balanceOf(self), assetPayout2, 1);
     }
 
+    function testMulticallingEnableLockDepositRequestAndExecute(uint256 amount) public {
+        amount = uint128(bound(amount, 4, MAX_UINT128));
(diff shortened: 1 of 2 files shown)


hieronx (approved):

test/integration/CentrifugeRouter.t.sol
```diff
@@ -270,6 +270,34 @@ contract CentrifugeRouterTest is BaseTest {
...
+        vm.label(vault_, "vault");
+
+        erc20.mint(self, amount);
+        erc20.approve(address(router), amount);
+        centrifugeChain.updateMember(vault.poolId(), vault.trancheId(), self, type(uint64).max);
+
+        // multicall
+        uint256 fuel = estimateGas();
+        bytes[] memory calls = new bytes[](2);
+        calls[0] = abi.encodeWithSelector(router.enableLockDepositRequest.selector, address(vault_), amount);
```
peculiarity, line 288: If this test will be used as an example for call encoding from the UI, it better uses `abi.encodeCall`.   `encodeCall` makes args type check.

AStox: Updated!

PR #416, opened 2024-10-02
Gg/request redeem router target

peculiarity:


test/recon-core/Properties.sol
```diff
@@ -303,7 +303,7 @@ abstract contract Properties is Setup, Asserts, ERC7540CentrifugeProperties {
                 == (
                     sumOfFullfilledDeposits[address(trancheToken)] + sumOfRedeemRequests[address(trancheToken)]
                         - sumOfClaimedDeposits[address(trancheToken)] - sumOfClaimedRedeemCancelations[address(trancheToken)]
-                        - sumOfClaimedRequests[address(trancheToken)]
+                        - sumOfClaimedRequests[address(trancheToken)] - sumOfRedeemRequestsRouter[address(trancheToken)]
```
hieronx, line 306: Why isn't this:

```suggestion
                        - sumOfClaimedRequests[address(trancheToken)] + sumOfRedeemRequestsRouter[address(trancheToken)]
```

Since `sumOfRedeemRequests` is also added, not subtracted?

peculiarity: Actually if you check latest commit, it is exactly like that.

hieronx (approved):

